PR #11, opened 2015-02-05
Add multiple targets

nagromc:
Hi @anishathalye,

Please review this pull request allowing users to define targets.

If you have multiple hosts with different configurations, it is useful to execute a consistent group of tasks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ without the annoyance of having to manually copy or link files.
 Dotbot itself is entirely self contained and requires no installation (it's
 self-bootstrapping), so it's not necessary to install any software before you
 provision a new machine! All you have to do is download your dotfiles and then
-run `./install`.
+run `./install -t <list of targets>`.
 
 Template
 --------
@@ -94,10 +94,10 @@ installer should be able to be run multiple times without causing any
 problems.** This makes a lot of things easier to do (in particular, syncing
 updates between machines becomes really easy).
 
-Dotbot configuration files are YAML (or JSON) arrays of tasks, where each task
-is a dictionary that contains a command name mapping to data for that command.
-Tasks are run in the order in which they are specified. Commands within a task
-do not have a defined ordering.
+Dotbot configuration files are YAML (or JSON) dictionaries of targets with
+arrays of tasks, where each task is a dictionary that contains a command name
+mapping to data for that command. Tasks are run in the order in which they are
+specified. Commands within a task do not have a defined ordering.
 
 ### Link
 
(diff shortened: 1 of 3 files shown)


anishathalye:
It's actually possible to achieve this effect in a variety of ways without adding additional complexity to Dotbot.

One way is the way I manage my dotfiles, with a common `dotfiles` repo and a `dotfiles-local` repo with branches for each machine (or category of config). I install both on every machine. I think this method works pretty well.

Another way to achieve an effect closer to what you have, with multiple targets that you can specify, is the following. You can have one file per target, `$TARGET.conf.yaml`, and then you can modify the `install` script to take targets on the command line and run all the corresponding installations. Here is an example install script:

```bash
#!/usr/bin/env bash

set -e

DEFAULT_CONFIG_PREFIX="default"
CONFIG_SUFFIX=".conf.yaml"
DOTBOT_DIR=".dotbot"

DOTBOT_BIN="bin/dotbot"
BASEDIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"

cd "${BASEDIR}"
git submodule update --init --recursive "${DOTBOT_DIR}"

for conf in ${DEFAULT_CONFIG_PREFIX} ${@}; do
    "${BASEDIR}/${DOTBOT_DIR}/${DOTBOT_BIN}" -d "${BASEDIR}" -c "${conf}${CONFIG_SUFFIX}"
done
```

With this, for example, you can have a `default.conf.yaml`, and say, a `mac.conf.yaml` and `work.conf.yaml`. You can install by running

```bash
./install mac work
```

And it'll execute all three things.


Unless there are additional reasons to add this complexity to Dotbot itself, I'd prefer to avoid the additional complexity.

dsifford:
@anishathalye First: Thanks for your hard work on this... Came across dotbot yesterday and I'm already feeling the benefits

Anyway: I thought I'd mention that I found your above comment extremely helpful. I'd suggest perhaps adding that to the README file. @nagromc and I can't be the only ones interested in different workflows / machine.

Thanks again :beers:

anishathalye:
Glad you're finding it useful!

Good point on adding it to the README. I'll re-open this issue, and I'll get to making the change as soon as I can.

anishathalye:
Oh whoops nvm, this is a pull request, not an issue. See #72.

ProfeC:
@anishathalye not 100% sure how my commit got to be in this pull request. I just added a link for reference in my commit. Please disregard.

_Sorry!_


... and **_THANKS_**!!! I'm really loving centralizing my config files with dotbot. Who knew!

anishathalye:
Oh, no worries!

I'm glad to hear that you've found the tool to be useful! 😄

borisrorsvort:
Dunno what’s wrong with the example above but I’ve not manage to make it work. I had to change it to: 
```
configs=("$DEFAULT_CONFIG_PREFIX $@")

for CONFIG in $configs; do
  "${BASEDIR}/${DOTBOT_DIR}/${DOTBOT_BIN}" -d "${BASEDIR}" -c "${CONFIGS_DIR}${CONFIG}${CONFIG_SUFFIX}"
done
```

Can you verify if that makes sense @anishathalye ?

anishathalye:
huh, not sure why the first one doesn't work for you…

dsifford:
The reason it doesn't work for @borisrorsvort might (emphasis _might_) be because the expansion isn't double-quoted in @anishathalye's example.

borisrorsvort:
After some more investigation it’s not executing the next files if they were errors in the default one

anishathalye:
Ah, okay... that's because of the `set -e` at the top of the script. See [here](https://www.gnu.org/software/bash/manual/bashref.html#The-Set-Builtin) for more info.

PR #27, opened 2015-04-24
Provide exception information to help debug issues when reading the conf

joshblum:
When getting started I had a few typos in my YAML something like this would have helped pinpoint the error quickly instead of digging into the source to print stuff to see where the issue was. 

We can also remove the `ReadingError` completely since it doesn't really provide any helpful information to fix the issue.

### DIFF
--- a/dotbot/config.py
+++ b/dotbot/config.py
@@ -9,8 +9,8 @@ class ConfigReader(object):
             with open(config_file_path) as fin:
                 data = yaml.load(fin)
             return data
-        except Exception:
-            raise ReadingError('Could not read config file')
+        except Exception as e:
+            raise ReadingError('Could not read config file:\n%s' % e)
 
     def get_config(self):
         return self._config


anishathalye:
Good idea!

I think the error message looks better if it's indented (kind of how stack traces look in Python).

How does this look to you? https://github.com/anishathalye/dotbot-dev/commit/337ede546077fe9c486c368f205aa68164e4a207

If that looks good, I can go ahead and merge it.

joshblum:
Awesome! left a comment but basically looks good :D

anishathalye:
I'm aware that Dotbot doesn't follow the convention with kwargs (see other files in the source)... perhaps it's time to switch and follow PEP 8.

I'll fix that and then merge this in.

anishathalye:
Thanks for the patch! Merged in 3a649c1.
You can see the changes in 56d8d05.

joshblum:
Awesome! Yeah I use flake8 for linting and autopep8 in my editor they work pretty well I think it's worth it for the change.

PR #30, opened 2015-04-24
Add pep8 conformity using flake8

joshblum:
Thought this might be nice :)

If you're interested, might be worth adding a [hook for commits](https://flake8.readthedocs.org/en/2.3.0/vcs.html) and maybe updating `CONTRIBUTING.md`?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,8 @@
+.PHONY: clean lint
+
+clean:
+	find . -type f -name '*.py[cod]' -delete
+	find . -type f -name '*.*~' -delete
+
+lint: clean
+	-flake8 .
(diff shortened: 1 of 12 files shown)


anishathalye:
I don't necessarily like all the things that flake8 enforces, such as lining up arguments in function calls.

Also, I don't like making whole-project style changes on this scale (it makes `git blame` more annoying to use, among other things), I think it's nicer to gradually change things as the code changes, so I think it's better to leave this as is.

joshblum:
Reasonable. If you want to change later you can ignore the function overhang stuff pretty easily with the `setup.cfg`

Issue #32, opened 2015-04-27
force doesn't work for symlinked directories


antaflos:
I am in the process of migrating my custom dotfiles install script to Dotbot, and my dotfiles from an old, defunct Mercurial repo to Github. This means that almost all dotfiles I want Dotbot to manage are already symlinked. Naturally I tell Dotbot to force creation of the new symlinks, but this doesn't seem to work for symlinked directories, like `~/.vim`:

```
$ ls .vim
lrwxrwxrwx 1 daff daff 28 Jan  1  2013 .vim -> /home/daff/dotfiles.hg/.vim/
```

My `install.conf.yaml` looks like this:

```yaml
- clean: ['~']

- link:
    ~/.dotfiles: ''
    ~/.tmux.conf:
      force: true
      path: .tmux.conf
    ~/.vim:
      force: true
      path: .vim/
    ~/.vimrc:
      force: true
      path: .vimrc

- shell:
  - [git submodule update --init --recursive, Installing submodules]
```

When running `./install` I get this:

```
All targets have been cleaned
Creating link ~/.dotfiles -> /home/rz01/daff/dotfiles/
Removing ~/.vimrc
Creating link ~/.vimrc -> /home/rz01/daff/dotfiles/.vimrc
Removing ~/.tmux.conf
Creating link ~/.tmux.conf -> /home/rz01/daff/dotfiles/.tmux.conf
Failed to remove ~/.vim
Incorrect link ~/.vim -> /home/rz01/daff/configs/.vim
Some links were not successfully set up
Installing submodules [git submodule update --init --recursive]
All commands have been executed

==> Some tasks were not executed successfully
```

The exception `linker.py` throws is `Cannot call rmtree on a symbolic link`. It seems the code in https://github.com/anishathalye/dotbot/blob/master/dotbot/executor/linker.py#L74 tries to delete the symlink after dereferencing it and thus tries to delete the target directory, in this case `/home/daff/dotfiles.hg/.vim/` which `shutil.rmtree` doesn't allow. Not sure if I interpreted that correctly.

A simple solution could look like this:

```
diff --git a/dotbot/executor/linker.py b/dotbot/executor/linker.py
index d94a9bc..9821fe7 100644
--- a/dotbot/executor/linker.py
+++ b/dotbot/executor/linker.py
@@ -78,7 +78,9 @@ class Linker(Executor):
                 (self._exists(path) and not self._is_link(path))):
             fullpath = os.path.expanduser(path)
             try:
[...]

PR #33, opened 2015-04-27
Handle force creation of symlinked directories

antaflos:
Check if entry to force-create (delete) is a symlink and if so, simply
unlink it. Don't try to run rmtree on the underlying directory, which
will fail anyway.

Addresses issue #32.


anishathalye:
LGTM!

Do you mind amending the commit message to remove the "Addresses issue #32."?
I prefer to keep issue references unidirectional -- I don't like referencing issue numbers in commit messages.

antaflos:
@anishathalye Done!

anishathalye:
Merged. Thanks for the patch!

Issue #23, opened 2015-03-25
Add support for the XDG Base Directory Speficiation


anishathalye:
I'm not sure how many people use this. It might be nice to add support for the [XDG Base Directory Specification](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) so files to link can be specified relative to `$XDG_CONFIG_HOME`.

I'm not sure what's the best interface for this -- perhaps doing a string replacement in the link target would suffice.

anishathalye:
Also, this is (partially) fixed in #37. It should be possible to follow the XDG spec by just using a `$XDG_CONFIG_HOME` in the link target now.

In Dotbot 2.0, it may be nice to add better support for this with nice syntactic sugar.

PR #37, opened 2015-05-07
Ability to use environment variables in link paths

PacketPerception:
Added the ability to use environment variables in link paths, and set up the HOSTNAME environment variable if it isn't already set. This makes linking host specific configs very easy by hostname, and leaves the potential to use other environment variables in the paths.

For example:
```yaml
- link:
    ~/.gitconfig: host/$HOSTNAME/gitconfig
```

Feel free to use or not, works well for my use case.

### DIFF
--- a/dotbot/cli.py
+++ b/dotbot/cli.py
@@ -1,3 +1,5 @@
+import os
+import socket
 from argparse import ArgumentParser
 from .config import ConfigReader, ReadingError
 from .dispatcher import Dispatcher, DispatchError
@@ -24,6 +26,8 @@ def read_config(config_file):
 
 def main():
     log = Messenger()
+
+    os.environ.setdefault("HOSTNAME", socket.gethostname())
     try:
         parser = ArgumentParser()
         add_options(parser)
--- a/dotbot/executor/linker.py
+++ b/dotbot/executor/linker.py
@@ -107,7 +107,7 @@ class Linker(Executor):
         Returns true if successfully linked files.
         '''
         success = False
-        source = os.path.join(self._base_directory, source)
+        source = os.path.expandvars(os.path.join(self._base_directory, source))
         if (not self._exists(link_name) and self._is_link(link_name) and
                 self._link_destination(link_name) != source):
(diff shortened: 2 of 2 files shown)


anishathalye:
Thanks for the patch. It'll probably be a couple days before I can take a look at this.

anishathalye:
Okay, turns out that I had some time today. A couple thoughts:

* I thought that `expandvars()` might expand unset variable names to the empty string, which could result in bad things like [this](http://www.slashgear.com/steam-on-linux-bug-can-delete-all-users-files-16364945/) happening. Fortunately, according to the [documentation](https://docs.python.org/2/library/os.path.html#os.path.expandvars), it leaves unset variables unchanged.
* This could make it possible to support the XDG base directory specification (see #23), although I guess it would be kinda nice to add syntactic sugar for that at some point.
* Setting the hostname seems like it shouldn't be something that is done in Dotbot itself. If someone wants to do that, they could add a line to the `install` script like `export HOSTNAME="$(hostname)"`.
* If we support variable expansion when linking, I think we need variable expansion in the `_delete()` method, linker.py:79 too.

One thing to consider:

* If we support variable expansion in the source, should we support variable expansion in the target as well? So then something like this would be okay: `~/$HOSTNAME: path/file`

PacketPerception:
Agreed with exporting $HOSTNAME in the install script, no idea why that didn't occur to me. I see no reason not to support expandvars in the source as well. Supporting in the delete method would be good as well.

anishathalye:
All right, I got around to implementing this: https://github.com/anishathalye/dotbot-dev/commit/5fcfde53ae31885025943ab2c9af0a17d3806bab

This required pretty minimal changes in the code (2 lines). I also added a couple tests for the feature.

I've credited you in the commit message. Let me know if it looks okay :)

PacketPerception:
Looks good to me! Thanks for the great project, it's been making my life a whole lot simpler.

anishathalye:
Okay, I'm going to push this (fd7f3b85 / tag v1.6.0).

Thanks for the idea for the patch!

PR #41, opened 2015-06-11
add init-dotfiles link

joshblum:
Add a link to the [easy-dotfiles](https://github.com/Aviator45003/easy-dotfiles) repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Get Running in 5 Minutes
 ### Starting Fresh?
 
 Great! You can automate the creation of your dotfiles by using amazing
-[easy-dotfiles][easy-dotfiles] script. Or, if you're just looking for [some
+[init-dotfiles][init-dotfiles] script. Or, if you're just looking for [some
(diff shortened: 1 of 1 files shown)


anishathalye:
I wonder if it would be better to add a separate "user-contributed" section to the README and put this in there (along with anything anyone else makes in the future).

Also, I'm going to wait till the name of the thing is finalized.

joshblum:
Yeah waiting for the name change is good, just wanted to put this in before I forgot. 

For this specific case I feel it makes more sense to put here since new users will otherwise clone the template repo and not use the automated script..

anishathalye:
I don't want to remove the link to the template. There may be people who are more comfortable with using something like that rather than a script.

anishathalye:
How's this: https://github.com/anishathalye/dotbot-dev/commit/dcc3a1947a85aaa038d143f9f6a0c2a1c1a76a15

joshblum:
Lgtm :)

anishathalye:
Okay, done!

PR #42, opened 2015-06-14
Realize not everybody has bash

Aviator45003:
OpenBSD does not come with `bash` by default. Rather than `pkg_add` it, I chose to correct the scripts so as to not require it when the scripts function just as well with `sh`.


Aviator45003:
Seems my eagerness was misplaced.

PR #43, opened 2015-06-14
Load configuration file in order

PacketPerception:
I have a use case where it is important that links are created in the order that listed, e.g.

```
- link:
    ~/.pyenv: contrib/pyenv
    ~/.pyenv/plugins/pyenv-virtualenv: contrib/pyenv-virtualenv
```
This way I don't have to modify the pyenv submodule to add pyenv-virtualenv as a submodule to it.

I've implemented this by having pyyaml load using an OrderedDict rather than a dict. I also added a test that I confirmed will fail with the default loading.

### DIFF
--- a/dotbot/config.py
+++ b/dotbot/config.py
@@ -1,5 +1,24 @@
 import yaml
 from .util import string
+from collections import OrderedDict
+
+
+def ordered_load(stream, Loader=yaml.Loader, object_pairs_hook=OrderedDict):
+    """ Stolen from SO - http://stackoverflow.com/questions/5121931/in-python-how-can-you-load-yaml-mappings-as-ordereddicts
+
+    Loads a YAML file using an OrderedDict so the ordering of the values are
+    maintained
+    """
+    class OrderedLoader(Loader):
+        pass
+    def construct_mapping(loader, node):
+        loader.flatten_mapping(node)
+        return object_pairs_hook(loader.construct_pairs(node))
+    OrderedLoader.add_constructor(
+        yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG,
+        construct_mapping)
+    return yaml.load(stream, OrderedLoader)
+
 
 class ConfigReader(object):
     def __init__(self, config_file_path):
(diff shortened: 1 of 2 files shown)


anishathalye:
Hmm, there is already a way to express order in the current syntax:

```
- link:
    ~/.pyenv: contrib/pyenv
- link:
    ~/.pyenv/plugins/pyenv-virtualenv: contrib/pyenv-virtualenv
```

It does involve repeating the "link", so that's slightly annoying, but it's possible. I don't know if there's enough of a benefit by using ordereddicts that it's worth doing this.

PacketPerception:
Yeah, I've recently come to realize this. I guess the question would be,
are there any downsides to using the ordered dicts then? It seems nice to
me to have a more intuitive set of actions take place than the somewhat
(seemingly) random order you get with the default behavior.

anishathalye:
It doesn't cause any breakage to use the given order (because it's one specific ordering of the things in the dictionary, and any ordering is acceptable). However, I think if we add this functionality, people will start relying on it, and then we'll be forced to maintain dealing with ordered dictionaries, having an alternative interpretation of JSON/YAML.

PacketPerception:
I'm not sure I follow on the different interpretation of JSON/YAML. Since JSON is a subset of YAML, the YAML loader does the job for both formats, so the same code works for both.

anishathalye:
By interpretation of JSON/YAML, I'm not talking about the code working properly to interpret stuff as ordered.

In the JSON / YAML spec, dictionaries are not considered ordered. Like the JSON `{"a": 1, "b": 2}` and the JSON `{"b": 2, "a": 1}` are describing the same structure. Interpreting the two JSONs differently seems to diverge from the JSON spec/semantics.

anishathalye:
I'm going to go ahead and close this. There is already a way of specifying the order in which to do things, so I don't think there's an advantage in using ordered dictionaries.

anishathalye:
I appreciate you taking the time to submit the patch.

By the way, in the future, you might save time if you open an issue to start a discussion before working on patch.

Issue #44, opened 2015-06-15
CLI flag to only run certain sections?


PacketPerception:
Would it make sense to add a flag to only run the clean, link, or shell sections of the config? Sometimes I dont want the shell commands to run so I just comment them out. This could be implemented using flags (dotbot -c/--clean -l/--link -s/--shell), or possibly as "sub-commands" (dotbot {clean,link,shell} or dotbot run {clean,link,shell}) and the default would still be to perform all three tasks.

I also thought it might be useful to add another optional argument to shell configurations to prompt the user if they want the shell command to run or not, something like "confirm: true". Not really sure that would be useful if I could just tell dotbot to only clean and link.

Aviator45003:
Problem is, sometimes people re-order certain commands to run other commands before. In more complicated setups, this means link commands, shell commands, then some link commands, are run in various orders, and things might fail if the shell commands before them were not run.

A possible solution for you would be to export your shell commands to scripts in your repository that, before they run, confirm you want to run.

PacketPerception:
I don't agree that your scenario would introduce any problems, as I don't think the default behavior should change at all. It would just be extra, optional, arguments. 

Currently, I'm doing what you suggested, creating external scripts that prompt for confirmation. This already requires settings extra arguments (stdin: true and stdout: true), so adding a confirm option would make the configuration file simpler and not force users to clutter there repos with extra scripts that simply confirm whether or not they want the command to run.

Aviator45003:
So it's not actually CLI arguments you are looking for, just an option in YAML to prompt before running?

PacketPerception:
For the confirm option yeah. I see how I should have made that more clear >D

Aviator45003:
I'll poke at the source code and see if I understand the structure enough to do that.

PacketPerception:
Nice of you to do so! I'd recommend looking at the executor/commandrunner.py script, adding a check for item.get('confirm', False) is True (to follow the convention already in place), then ask for confirmation and just ``continue`` to the next command if no, otherwise run it.

anishathalye:
What's the use case for this? What kinds of commands might you want to only run sometimes? And when would you want to leave the decision up to the user instead of automating the decision making (e.g. rewriting "fetch and install app" to an idempotent operation like "if application is not installed, fetch application from internet and install it").

[...]

PR #45, opened 2015-06-19
Address #44

Aviator45003:
@PacketPerception, mind testing this PR, and confirm it addresses #44 adequately?


Aviator45003:
@anishathalye Do you want me to break the `_query_yes_no` out into its own file? How about input, does that matter anywhere except here?

anishathalye:
Going to address #44 in a different way (see https://github.com/anishathalye/dotbot/issues/44#issuecomment-115105960).

Issue #40, opened 2015-06-10
Test using both python2 and python3


anishathalye:
Dotbot, at least with `1.x`, will maintain compatibility with both Python 2 and Python 3.
This can sometimes be painful to do, and sometimes, it can be error-prone. In the past, there have been bugs with cross-version compatibility.

It would be good to modify the testing code so that tests are run with both Python 2 and Python 3. Currently, it only tests with Python 2.

PacketPerception:
While the bash testing is elegant and simple, would it be more beneficial to move to a testing suite that could be run in something like Travis CI? Thinking python unittest test cases running within temporary directories. Tests would run a lot faster as well.

anishathalye:
Yeah, it would be awesome if we could move to a testing suite that could be run in Travis CI.

I'm not sure how unit test cases running within temporary directories would work - I wouldn't want a broken test case or broken code to screw up a machine. That's why I went with the whole VM thing (that's super duper slow...).

I've spent a tiny amount of time looking into file system mocking in Python, but I haven't read enough about it to figure out if it's a good choice for what we're doing.

Figuring out how to make tests run faster (in the VM or outside the VM) would be awesome.

anishathalye:
It looks like this might be a reasonable way of testing with multiple Python versions: http://redsymbol.net/articles/env-and-python-scripts-version/

PacketPerception:
When testing with Travis-CI, this is a non-issue, check out my travis configuration for one of my projects: https://github.com/PacketPerception/pychievements/blob/master/.travis.yml

That results like so: https://travis-ci.org/PacketPerception/pychievements and even generates coverage reports via coveralls, https://coveralls.io/builds/2864248

Here I'm using nose which is just a super simple way of running unittest.TestCase tests. I'd be more than happy to try write one up as an example to see if it would work out. 

For testing locally on your own machine, I highly recommend pyenv for managing your python versions, then using tox to automatically run tests against multiple versions. See http://tbb.co/using-tox-and-pyenv-for-testing-and-development/ for an example

anishathalye:
Yeah, good point about FS modification and bugs not being an issue with Travis-CI. Coveralls is nice too.

If it wouldn't be too much work, it would be nice to see what a test case would look like.

And thanks for the link about pyenv/tox. I didn't know too much about good ways to do that.

PacketPerception:
[...]

PR #46, opened 2015-06-28
Convert tests to use Python unittest with tox/pyenv

PacketPerception:
See #40


anishathalye:
Thanks for the pull request. This is a really busy week for me; I'll take a look as soon as I can.

anishathalye:
I think it's dangerous to have tests use the real file system - if there's a bug in Dotbot code, something bad could happen. The current approach (with the VM) is slow, but it works, and the tests only need to be run every time there's an update (which doesn't happen all that often), so I don't think it's a huge problem. So for now, I think we'll stick with the old format of the tests.

PR #58, opened 2015-12-07
Add Dirrer for creating directories

casey:
I added a dir creating directive. I like it since I have a lot of files I need to link into `bin` and `.vim`.

### DIFF
--- a/dotbot/executor/__init__.py
+++ b/dotbot/executor/__init__.py
@@ -2,3 +2,4 @@ from .executor import Executor
 from .linker import Linker
 from .cleaner import Cleaner
 from .commandrunner import CommandRunner
+from .dirrer import Dirrer
--- a/dotbot/executor/dirrer.py
+++ b/dotbot/executor/dirrer.py
@@ -1,0 +1,46 @@
+import os
+from . import Executor
+
+class Dirrer(Executor):
+    '''
+    Creates directories.
+    '''
+
+    _directive = 'dir'
+
+    def can_handle(self, directive):
+        return directive == self._directive
+
+    def handle(self, directive, data):
+        if directive != self._directive:
+            raise ValueError('Dirrer cannot handle directive %s' % directive)
+        return self._process_dir(data)
+
+    def _process_dir(self, targets):
+        success = True
+        for target in targets:
+            success &= self._dir(target)
+        if success:
+            self._log.info('All directories have been created')
+        else:
+            self._log.error('Some directories were not successfully created')
+        return success
+
+    def _dir(self, target):
+        '''
+        Creates directories, including intermediate directories.
+        '''
+        expanded_target = os.path.expanduser(target)
+        
+        if os.path.isdir(expanded_target):
+            return True
+
+        if os.path.isfile(expanded_target):
+            return False
+
+        try:
+            os.makedirs(expanded_target)
+        except:
+            return False
+
+        return True


anishathalye:
Is this necessary when link supports a [create](https://github.com/anishathalye/dotfiles/blob/e19e3c756d76276534e175ec5da19ebb01c9df95/.install.conf.yaml#L13) option?

```yaml
- link:
    ~/.sbt/0.13/build.sbt:
      create: true
      path: sbt/0.13/build.sbt
```

Also, I've just been [shelling out to mkdir](https://github.com/anishathalye/dotfiles/blob/e19e3c756d76276534e175ec5da19ebb01c9df95/.install.conf.yaml#L26) to create other directories that I've needed.

```yaml
- shell:
  - mkdir -p ~/.rtorrent/session ~/.rtorrent/watch ~/.rtorrent/downloads
```

casey:
It's just a convenience function, but I think it's worth it. In particular, it's nice to just be able to use the short form for links, and having to use the long form for a lot of links is much more verbose.

anishathalye:
There will be a way to have defaults in Dotbot 2.0, so you can just set `create: true` as a default for linking.

casey:
I think this could go either way, so I defer to your judgement.

A really interesting thing to do would be to look at people config files and see what common patterns are, and then try to support them better.

This search brings up tons of interesting results: https://github.com/search?utf8=%E2%9C%93&q=filename%3Ainstall.conf.yaml&type=Code&ref=searchresults

Do you know a way to do that search but return a list of filenames, so you could write a script to curl everyones config files, for further analysis?

anishathalye:
Okay, let's pass on it for now.

Haha, I've been doing something similar to that for some time (using this search: https://github.com/search?q=DOTBOT_DIR&ref=searchresults&type=Code&utf8=%E2%9C%93). I have cloned ~600 repos on my computer (using a script to grab the search results and clone them).

It would be really interesting to look through the install.conf.yaml files for common patterns to see what could be improved / made easier.

casey:
Could you put the script up somewhere? It could probably be adapted to extract conf files. (Or download them directly via curl, to avoid cloning the whole repo.)

anishathalye:
https://gist.github.com/5278586119d32a0612da

(this prints the names, but it should be easy to modify it to download conf files)

casey:
Dotbot user study in progress: github.com/casey/dotbot-user-study

casey:
I got distracted before I could do much with the config files, but if you're interested, they're here: https://github.com/casey/dotbot-user-study/tree/master/configs

anishathalye:
Cool! I might look into doing a more detailed user study eventually.

Also, please let me know if you end up doing more analysis!

casey:
Was just [updating my dotfiles to pull in the latest version of dotbot](https://github.com/casey/dotfiles/commit/407d6f6cbad82abd117bab18a297fa40e9871ccf) and was reflecting on the fact that I've been using dotbot for almost 7 whole years. It's been working flawlessly the whole time, and I've used it to bootstrap my dotfiles on countless laptops, pcs, and servers.

So here's to 7 more years (at least) and thanks for such a useful tool!

casey:
Also, I just commented here because it was the first interaction I had with the project, but I notice that dir creation was implemented, so [I can now use the `create` plugin](https://github.com/casey/dotfiles/commit/99f75b43fd65a51d82d3d4c67236671bec6b101b). Pretty sweet!

anishathalye:
Thank you for the kind words :smile:

Issue #59, opened 2015-12-15
Support ANSI sequences under windows


brandto:
dotbot works even under windows "as is" if you are able to create symlinks. So using python3 and having the rights to create symlinks you'll be able to use this great tool.

But as "cmd.exe" doesn't support ANSI you will not get colored output. Should be easy to fix using "colorama".

anishathalye:
Hmm I haven't used it Windows. Is the output just not colored, or does it have funny escape sequences in it?

Also, I don't want to add an external dependency. Hopefully, it shouldn't be hard to reimplement the functionality.

anishathalye:
I haven't used Dotbot on Windows before. How does this currently look?

I would prefer to avoid adding more external dependencies for something like this.

brandto:
As a matter of fact you'll get the 'ugly escape sequences'.

brandto:
Would it be ok for you if we just add a single file?! As an example https://github.com/pycontribs/tendo/blob/master/tendo/ansiterm.py could be used.

Another possibility: Check for installed colorama (or ansiterm) on the machine (if running under windows). If found something useable: use it. If not: Use "empty" ANSI sequences.

anishathalye:
Hmm I'm not sure if I want to include a couple hundred lines of code for prettier colors on Windows (also, the colors aren't _that_ useful).

Stripping escape sequences on Windows seems reasonable, though.

PR #60, opened 2015-12-15
Support ANSI sequences under windows

brandto:
Use "colorama" under windows in order to get colored output (Issue #59).


anishathalye:
See #59 for discussion.

Issue #67, opened 2016-01-15
Add continuous integration


anishathalye:
So things like #66 don't happen.

Looks like it's possible: http://www.lewuathe.com/blog/2015/01/25/vagrant-in-travisci/

CC @bchretien

bchretien:
cc @PacketPerception

Nice! Just a quick note: the link you posted presents an approach that depends on `sudo`, thus preventing the use of the [new container-based infrastructure](https://docs.travis-ci.com/user/workers/container-based-infrastructure/). This is not a requirement, but with it builds start faster (and availability is probably improved). It may not be possible yet though (apt packages need to be [vetted](https://github.com/travis-ci/apt-package-whitelist) etc.). Moreover, the person who wrote that post [no longer uses that](https://github.com/Lewuathe/storm-cookbook/commit/53b5b1c62b2a3480bb01b3faef236f00cacdfb7e) (nor Vagrant it seems, it's hard to tell). Then again, I haven't ported all my projects to the new infrastructure either for the exact same reasons, and it works just fine.

To investigate Travis CI, you should consider making a new branch (e.g. `travis`), and enable it on Travis (see [this](https://docs.travis-ci.com/user/customizing-the-build/#Building-Specific-Branches)). That way you don't pollute `master`, and you can just rebase the branch on `master` once things work as expected.

bchretien:
Actually, it may be even better to run the tests without Vagrant completely, since there is no risk of breaking the system. In that case, things are much easier, [this guide](https://docs.travis-ci.com/user/languages/python) shows the possibilities for Python projects.

anishathalye:
Fixed in #68.

PR #68, opened 2016-01-16
Add Travis CI support

bchretien:
This implements Travis CI support (see #67).

- Use the current Bash test suite.
- Add the possibility to run the test suite without Vagrant. Since on Travis, breaking the system is not an issue, I use a cleanup function instead. It merely removes files created after the start of the test suite. Since there is nothing besides the tests happening on the system, it should be good enough. Of course, that's something that should not be done on an actual system...
- Use the container-based infrastructure (`sudo` is not needed).
- Example of successful build available [here](https://travis-ci.org/bchretien/dotbot/builds/102755162). I also made a dummy commit breaking a test [there](https://travis-ci.org/bchretien/dotbot/builds/102755980).
- I haven't tested the Vagrant build locally with this commit, so it should probably be done before merging.

### DIFF
--- a/.travis/run_tests
+++ b/.travis/run_tests
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+set -e
+
+# For debug only:
+# export DEBUG=true
+# set -x
+# set -v
+
+BASEDIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+
+# Travis runs do not rely on Vagrant
+export USE_VAGRANT=false
+export DOTBOT_EXEC="${BASEDIR}/bin/dotbot"
+
+cd "${BASEDIR}"
+. "test/driver-lib.bash"
+
+travis_initialize() {
+    echo "initializing."
+    tests_run=0
+    tests_passed=0
+    tests_failed=0
+    tests_total="${1}"
+    local plural="" && [ "${tests_total}" -gt 1 ] && plural="s"
+    printf -- "running %d test%s...\n\n" "${tests_total}" "${plural}"
+}
+
+travis_cleanup() {
+    # Remove all dotfiles installed since the start, ignoring the main
+    # dotfiles directory, and the dotbot source directory
+    find ~ -mindepth 1 -readable -newermt "${date_stamp}" \
+        -not \( -path ~ -o -path "${BASEDIR}/*" \
+                -o -path ~/dotfiles \) \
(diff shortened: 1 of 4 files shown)


anishathalye:
That was quick. Thanks!

I don't like mentioning issue numbers in Git commits -- would you mind swapping that for the first three bullet points in your comment here?

I will test the build on my system before merging.

.travis/run_tests
```diff
@@ -0,0 +1,75 @@
...
+    echo "initializing."
+    tests_run=0
+    tests_passed=0
+    tests_failed=0
+    tests_total="${1}"
+    local plural="" && [ "${tests_total}" -gt 1 ] && plural="s"
+    printf -- "running %d test%s...\n\n" "${tests_total}" "${plural}"
+}
+
+travis_cleanup() {
```
anishathalye: Nice, this is a clever way to do it!

bchretien:
Reference to the GH issue removed from the commit message.

README.md
```diff
@@ -1,6 +1,8 @@
 Dotbot
 ======
 
+[![Build Status](https://travis-ci.org/anishathalye/dotbot.svg?branch=master)](https://travis-ci.org/anishathalye/dotbot)
```
anishathalye: Could you move this down a little bit (perhaps after the `---` separator)? It's not the most important thing for new users to see (ideally, the build should always be passing :stuck_out_tongue:)

bchretien: Sure!

.travis/run_tests
bchretien: Clearly not as foolproof as the Vagrant approach, but in that specific context, that seems to be good enough. A more complex method could be used if it was really necessary (e.g. taking snapshots of the home directory).

anishathalye:
Thanks so much for the patch! (also, that was super quick!!)

I tested locally, and everything is working great with Vagrant. I also ran tests with Travis-CI, and everything is looking great there too.

Merged into master in 2e64388.

bchretien:
Great, thanks!

bchretien:
I noticed that apparently, while the builds on my fork ran on the container-based infrastructure, dotbot builds used the legacy infrastructure. I guess we should [explicitly set that](https://docs.travis-ci.com/user/migrating-from-legacy/#How-can-I-use-container-based-infrastructure%3F) in `.travis.yml`.

PR #69, opened 2016-01-17
Travis: minor improvements

bchretien:
This PR contains minor improvements for #68.

- Force the use of the container-based infrastructure. According to their doc, I assumed that would be the case by default, but apparently not:

> For repos we recognize before 2015-01-01, linux builds are sent to our standard infrastructure.
> For repos we recognize on or after 2015-01-01, linux builds are sent to our container-based infrastructure.

- Since the test script for Travis is now in the `test` directory, I added a check to prevent running it outside of Travis. It depends on [their environment variables](https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,5 +6,7 @@ python:
   - "3.4"
   - "3.5"
 
+sudo: false
+
 script:
   - ./test/test_travis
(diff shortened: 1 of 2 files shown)


anishathalye:
Looks great. Minor nit: could you use <= 50 char commit messages (see http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html)

bchretien:
Ok, commits reworded ;)

PR #71, opened 2016-02-06
Directive plugins

casey:
Okay, after much delay, I have a PR for you that implements plugins.

I've tested it on 2.7, 3.2, 3.3, 3.4, and 3.5. I suspect it also works on 3.0 and 3.1.

I turned shell, link, and clean into plugins. So that the built in plugins are easy to find and understand, to encourage users to copy them and make their own, I put them in a top-level plugins folder, and renamed them so that they're just their directive name.

I changed the name of Executor to Plugin. It can be used from plugins as `from dobot import Plugin`, or `import dotbot` and the referred to as `dotbot.Plugin`.

Let me know if you'd like to me to make any changes.

### DIFF
--- a/dotbot/cli.py
+++ b/dotbot/cli.py
@@ -1,8 +1,11 @@
+import os, glob
+
 from argparse import ArgumentParser
 from .config import ConfigReader, ReadingError
 from .dispatcher import Dispatcher, DispatchError
 from .messenger import Messenger
 from .messenger import Level
+from .util import module
 
 def add_options(parser):
     parser.add_argument('-Q', '--super-quiet', dest='super_quiet', action='store_true',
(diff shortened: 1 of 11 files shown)


dotbot/cli.py
```diff
@@ -28,12 +37,24 @@ def main():
...
+          plugin_directories.append(os.path.join(os.path.dirname(__file__), '..', 'plugins'))
+        plugin_paths = []
+        for directory in plugin_directories:
+          for plugin_path in glob.glob(os.path.join(directory, '*.py')):
+            plugin_paths.append(plugin_path)
+        for plugin_path in options.plugins:
+            plugin_paths.append(plugin_path)
+        for plugin_path in plugin_paths:
+            abspath = os.path.abspath(plugin_path)
+            module.load(plugin_path)
```
anishathalye: `abspath` isn't used here. Bug? Or does this work fine, and we can get rid of line 56?

dotbot/util/module.py
```diff
@@ -0,0 +1,29 @@
...
+loaded_modules = []
+
+def load(path):
+  basename = os.path.basename(path)
+  module_name, extension = os.path.splitext(basename)
+  plugin = load_module(module_name, path)
+  loaded_modules.append(plugin)
+
+if sys.version_info >= (3, 5):
+  import importlib.util, os.path
```
anishathalye: Duplicate import of `os.path` in this code path.

anishathalye:
Looks great! There were a couple minor things I noticed when I read the code, so I left comments on a couple lines.

After that is fixed, I should be able to merge this in.

anishathalye:
Also, could you rebase the single commit onto the current HEAD? I don't want 1afc8bf.

anishathalye:
By the way, I added some documentation to the readme and wrote a couple really simple tests, so I'll add that commit on top of yours.

casey:
Heyo, I noticed you fixed up the issues and merged it in. Awesome ^_^ I use abspath as the path to load modules so that when you print out a module or there's an error involving a module, the message has a readable path.

anishathalye:
Yep, I was just about to comment on this. Merged eeb4c28 in aaf093b, added docs/tests in c402396.

Thanks so much for the contribution!

Issue #35, opened 2015-05-01
Dotbot 2.0


anishathalye:
This issue is meant for tracking work on Dotbot v2.0, the first major release since v1.0 in March 2014. All features that are backwards compatible are being included in Dotbot v1.0 as they are added. We'll only switch to the name "v2.0" if/when we decide to add features that are not backwards compatible.

**Completed features**
- Directive plugins (see #65)
- Default options (see #34 and #77)

**Accepted features / changes**
- Improved error messages (see #73)
- Backup (see #90)

**Features under consideration**
- CLI flag to only run certain sections (see #44 and #49, 2nd item)
- Platform specific link paths (see #57)
- Be able to write to files/folders that require special privileges (see #49, 3rd item)
- Uninstall command to delete symlinks specified in install config (see #49, 4th item)
- Automatic backup / replacement of dotfiles (see #12 and #36)
- Support interactive execution (see #74)
- Make Dotbot usable as a Python library (see #75)

**Rejected features**

(none)

---

**Design decisions**

(nothing here yet)

**To be discussed**
- What file format do we want to use? There are a couple options I can think of: JSON/YAML file (declarative), other data format (e.g. HOCON), or Python DSL. Maybe there are other completely different ways to do it too.
- Because people use Dotbot as a submodule, we need to be careful when figuring out how to ship the upgrade. Should it be on the `master` branch? Should Dotbot 2.0 forever live in a `2.0` branch with `master` being reserved for bugfixes for 1.0?

---

**Release information**

The release date is currently unknown. Work on Dotbot v2.0 has not started yet.

(This top post will be edited as stuff happens. This issue will be used for discussing the design of v2.0.)

anishathalye:
There may be some config file API changes required to have a clean way to support the XDG Base Directory Specification. See #23.

anishathalye:
It would be great to provide better error messages when there are problems with the config file. Currently, if there are YAML syntax errors, the error messages are decent. If the config file is valid YAML but doesn't make sense for Dotbot, the errors are pretty bad. If there is going to be a new config file format for Dotbot 2.0, this should be kept in mind so that we can be sure to have high-quality error messages. See #8 and #27 for related discussion.

anishathalye:
In the current config file format, there isn't a clear place to put default options. This seems like it is a nice feature to have, and this should be kept in mind if designing a new config file format. See #34.

joshblum:
+1 for the error messages

anishathalye:
[...]

PR #77, opened 2016-02-27
[RFC] Add default options

aykamko:
Adds default options as discussed in #35 

`defaults` didn't seem like a plugin-able action to me, so I wrote it as a special case.

### DIFF
--- a/dotbot/dispatcher.py
+++ b/dotbot/dispatcher.py
@@ -5,6 +5,7 @@ from .messenger import Messenger
 class Dispatcher(object):
     def __init__(self, base_directory):
         self._log = Messenger()
+        self._defaults = {}
         self._set_base_directory(base_directory)
         self._load_plugins()
 
@@ -20,11 +21,16 @@ class Dispatcher(object):
         success = True
         for task in tasks:
             for action in task:
+                if action == 'defaults':
+                    self._defaults = task[action]
+                    self._log.info('Set defaults')
+                    continue
                 handled = False
                 for plugin in self._plugins:
                     if plugin.can_handle(action):
                         try:
-                            success &= plugin.handle(action, task[action])
+                            success &= plugin.handle(action, task[action],
+                                                     self._defaults.get(action, {}))
                             handled = True
                         except Exception:
                             self._log.error(
@@ -36,8 +42,7 @@ class Dispatcher(object):
         return success
 
     def _load_plugins(self):
-        self._plugins = [plugin(self._base_directory)
-            for plugin in Plugin.__subclasses__()]
+        self._plugins = [plugin(self._base_directory) for plugin in Plugin.__subclasses__()]
 
 class DispatchError(Exception):
     pass
--- a/dotbot/plugin.py
+++ b/dotbot/plugin.py
@@ -15,7 +15,7 @@ class Plugin(object):
         '''
         raise NotImplementedError
 
-    def handle(self, directive, data):
+    def handle(self, directive, data, defaults):
         '''
         Executes the directive.
 
--- a/plugins/clean.py
+++ b/plugins/clean.py
@@ -10,7 +10,7 @@ class Clean(dotbot.Plugin):
     def can_handle(self, directive):
         return directive == self._directive
 
-    def handle(self, directive, data):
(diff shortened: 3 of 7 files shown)


anishathalye:
I like it! Let's leave the discussion open for a little bit before merging so we can get feedback from others like @casey who were involved in the discussion about this.

Once that's done, we should add to the documentation in the readme as well, and we should probably add a test case.

aykamko:
@anishathalye sounds good!

casey:
Looks good!

I added a comment about doing some extra validation on the defaults object. It would be a good opportunity to print a helpful error message, and hopefully make it easier for users to debug config files, for example if they misspell a directive.

It looks like with this implementation we support multiple defaults blocks, with the semantics that each defaults block overrides the previous one. I think this is reasonable, but we should definitely call that out in the readme so that people know that order matters.

Another thing to think about is that, currently, plugins are able to implement the "can_handle(action)" method, which gives them considerable flexibility in which actions they can handle. They could, for example, handle any action that matches a regex, for example "remote-shell-[a-z0-9]+" and then run a command on a remote host with that name. I don't think anyone is taking advantage of this, but the possibility is there.

Given that, there's a little bit of a mismatch between how defaults get passed to plugins. `self._defaults.get(action, {}))` will do exact string matching, so a plugin that accepts patterns will only get the defaults value which matches the exact string that it is currently handling.

I don't think this really matters, since no plugin I know of actually accepts patterns, but it's good to keep in mind.

aykamko:
@casey Yeah, I was actually thinking that passing `defaults` in to `handle()` seemed a little too restrictive. For example, what if some particular plugin wanted to reference defaults from another plugin? The way I set it up in this PR wouldn't allow for that.

I think a better solution might be to pass a reference to a global `plugin_defaults` dict when we `__init__` each plugin in `_load_plugins`. This would solve both the problem @casey poses and the one I just mentioned. As an added precaution, we could make `plugin_defaults` immutable for plugins. It would only be mutable through the special `defaults` directive handled by the `Dispatcher`.

@anishathalye Thoughts? This would be a slight restructure to the plugin interface.

anishathalye:
Random thought -- I don't know if this is actually a problem, but this will break old plugins (function arity mismatch). Not sure if we consider the current plugin interface stable (can this not land in v1.x / do we need a major version bump because we're breaking backwards compatibility).

I like supporting multiple `- defaults` directive (it seems more reasonable than having a somewhat arbitrary restriction that it can be done only once). Every time we specify another defaults set, we could either update (merge dicts) or replace. I like the idea of building a solution that allows for plugins to peek at the entire defaults dictionary for the reason that @casey mentioned.

Not sure exactly how the design should look (additional arg vs reference to global plugin_defaults dict). Another way is to add some object to manage this using a new `Context` object. So like the Plugin's current ivar `base_directory` would instead be accessed through a reference to a context object that the plugin has, `self.context.base_directory()`, and we could also have a method to get defaults like `self.context.defaults()`. Just another way of doing it.

I'm not sure if this is actually a good design. I'm kinda sleep-deprived at the moment, so I'll think about it again tomorrow when I'm more awake.

casey:
I think between merging subsequent defaults blocks and replacing them, I like replacing better, just because it's easier to keep track of, and there will probably be fewer surprises for users.

anishathalye:
Okay, makes sense, +1 for replacing rather than updating.

anishathalye:
Okay, I implemented a modified version of @aykamko's contribution: https://github.com/anishathalye/dotbot/commit/4f8a1cd6b7e8e1385489174695f9927954416a6e (with documentation and tests; credits in commit message).

It's currently on the develop branch, not merged into master.

How does it look?

anishathalye:
@casey thanks for the quick feedback on the patch!

I'll merge this in a day or two if there's no other feedback from anyone.

(I realize that this is slightly breaking the plugin API but I don't think it's a big deal, I don't think anyone is actually using plugins, and if they are, it should be a one minute fix.)

anishathalye:
Ok, made some finishing touches and merged in f52bbd1eec0d6fb103c5fcbb75c03ca223fffdb7.

If there's anything else that needs to be changed, the change can be applied on top.

Issue #73, opened 2016-02-21
Improve error messages


anishathalye:
cc @MaikuMori

anishathalye:
[Moving discussion here from #35. @darksider3]

Uhh, I haven't thought too much about this / haven't had too much time. If you want, feel free to start thinking about this and replying here with your thoughts. Or if you want, start hacking on it and see where it goes.

Darksider3:
Hey, just started Hacking on. 
How about adding errno from the built-in Python lib to match the Errno-codes as a "dependency"?
(E.g)
```python
Import errno
try:
...
Linking stuff
...
catch OSerror as err:
  if err.errno == errno.$SOME_EXPRESSIONS
     print " something usefull about fixing error based on errno-codes"
```
(Written from mobile, excuse if the codeblack isn't such fancy)

anishathalye:
Hmm, is it better to merely point out what caused the error (e.g. by using https://docs.python.org/2/library/os.html#os.strerror), or should we try to suggest fixes?

I'm not sure if I like suggesting fixes.

(Haha, I didn't know that GitHub did that either! Interesting feature...)

Darksider3:
Hm, yeah. Strerror should be enough :-) 
Im pretty sure that I don't like it. But that was the example how the entire "improvement" would look like. For sure I can easily replace it with the Error-String.
```python
print "Error: %s" %err.strerror
```
That would be enough i think. 
(YEPP, an night sleeping over it changed the mind..^^)

anishathalye:
Also, using strerror allows us to be lazier :P

Darksider3:
+ independent Language(In Linux it was possible to change error-locales provided by the kernel/Bash by installing locales wasn't it?) +Portability(as Windows 10 looks like giving an interface to ELF in the last days)

anishathalye:
Closing due to age / inactivity.

PR #84, opened 2016-04-01
Improve Error-Messages in link.py

Darksider3:
... by giving os-provided error messages to the output for linking. As discussed in #73 and #35 .

### DIFF
--- a/plugins/link.py
+++ b/plugins/link.py
@@ -74,7 +74,7 @@ class Link(dotbot.Plugin):
                 os.makedirs(parent)
             except OSError as err:
                 self._log.warning('Failed to create directory %s' % parent)
-                self._log.warning('Error: %s.' % err.strerror)
+                self._log.debug('Error: %s.' % err.strerror)
                 success = False
             else:
                 self._log.lowinfo('Creating directory %s' % parent)
@@ -100,7 +100,7 @@ class Link(dotbot.Plugin):
                         removed = True
             except OSError as err:
                 self._log.warning('Failed to remove %s' % path)
-                self._log.warning('Error: %s' %err.strerror)
+                self._log.debug('Error: %s' % err.strerror)
                 success = False
             else:
                 if removed:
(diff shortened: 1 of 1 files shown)


Darksider3:
What is about clean.py? Should we add there to function _in_directory and _clean some exceptions(esp. on removing symlinks?).
It's possible that the User has changed something and the permissions was changed during his last install of files(or got error messages even by installing before)..

anishathalye:
Hmm, maybe we should only show these things when the user passes `-v` using a lower log level (info or debug or something).

Darksider3:
U surely mean Log(Log.WARNING, "Error %s:" % err.strlen) ?

anishathalye:
`self.log.debug()` or `self.log.lowinfo()`, so that it only shows that information if people want to see it.

Darksider3:
And now, what about the clean.py?

Darksider3:
How about that? :-)

anishathalye:
I know it's a small diff, but I've been kinda busy... will review this as soon as I can.

MaikuMori:
Looks fine to me. I'm not so sure about hiding it behind debug flag. Error already is unexpected state, generally when I have error I also want the information to fix it.

Otherwise it's like just printing "ERROR", I'm looking at you Microsoft.

Darksider3:
First commit had it on error-level, debug ID in my opinion like..m "the Specific Programm wont work. Where is the failure in this Programm? Print me more about it".
An error given by a user would be IMO not about Debugging a Programm, more about debug the configuration file written by the user.

If you hide these messages, i think there would be Chance that some User might dont understand or know to use the debug Flag for this.
Another side, are those Users your adressed Group?

bbenne10:
I'm currently integrating dotbot into my dotfiles repo at home, and I would cast a vote that these error messages should be exposed by default. Even if they're printed directly after the "not handled" message versus on their own lines (which would be pretty trivial if we could stick with one version of python, but more difficult if we are targeting 2.7 and 3.x)

Looking over the changes, there are places where we're failing that I don't think we need to. For instance, line 75 is unconditionally setting success to False when a call to `os.makedirs` fails. I think we should probably check `err.errno` against `errno.EEXISTS` to see if the parent directory already exists and set success to `True` in this case. I'll play with this idea (once I get my zsh config linked properly -_-) and open a PR if I find the behavior to be more consistent. We can continue discussion there.

Darksider3:
If Line 75 fails because the User got not enough permissions, the OSerror is triggered. Otherwise it shouldn't throw the error, i thought?
Testung it today in a chroot.

PR #86, opened 2016-04-05
Fix linking when using both relink and relative

### LINKED ISSUE #85: Using relink or force with relative always removes before linking
The following config results in deleting and relinking the file every time the script is run:

```yaml
- defaults:
    link:
      relink: true
      relative: true

- link:
    ~/.ssh/conf.d/foo.conf:     ssh/conf.d/foo.conf
```

See https://github.com/anishathalye/dotbot/issues/70#issuecomment-205488166 for more information
Reported by @sjthespian

anishathalye:
Prior to this patch, the following config led to incorrect behavior:

    - link:
        ~/.folder/file:
          path: file
          create: true
          relative: true
          relink: true

Prior to the change, running the config the first time would result in
expected behavior. However, running the config for the second time would
result in deletion and re-creation of the link (even when the link is
correct).

This patch improves the interaction of relink and relative, taking
relative paths into account when checking the validity of existing links
in the `_delete()` method.

This fixes #85.

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,7 +294,7 @@ License
 Copyright (c) 2014-2016 Anish Athalye. Released under the MIT License. See
 [LICENSE.md][license] for details.
 
-[init-dotfiles]: https://github.com/Aviator45003/init-dotfiles
+[init-dotfiles]: https://github.com/Vaelatern/init-dotfiles
 [dotfiles-template]: https://github.com/anishathalye/dotfiles_template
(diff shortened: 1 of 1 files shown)


anishathalye:
This patch is slightly broken in its current state (it's still failing tests). I'll try to work more on it soon.

Also, a new unit test should probably be added for this.

anishathalye:
What... GitHub closed this pull request cause I reset the `develop` branch back to master and made a separate branch for this. Oh well, I'll open a new PR when I have a fixed commit.

PR #100, opened 2016-10-19
Use current $SHELL for shell-plugin, fix #97

### LINKED ISSUE #96: Feature request: add optional tasks
I would like Dotbot to install the required softwares by my dotfiles. For instance, if I use `i3`, I would add to `install.conf.yaml`:

``` yaml
- shell:
    - sudo apt-get install -y i3

- link:
    - ~/.config/i3/config: i3/config
```

But imagine I need multiple softwares like `fish`, `tmux`, Vim, etc. and with multiple package managers (`npm install -g bower gulp yo`, `sudo pip install pre-commit pre-commit-hooks thefuck`, etc.), each time I run `./install`, the whole installation process will take several minutes to run, especially on low resource systems such as the Raspberry Pi v1.

I know Dotbot should be [idempotent](https://github.com/anishathalye/dotbot/blob/28959a3f31d8632975de4e18d0a0e47476b5a413/README.md#configuration), but wouldn't be great to run `./install --init` or something to execute specific part of the file (for instance, install the required softwares)? Then, the `install.conf.yaml` would become:

``` yaml
- optional:
    # Only executed if `--init` is specified
    - name: init
      tasks:
        - shell:
            - [sudo apt-get install -y i3, Installing required softwares]

# Executed in any case
- link:
    - ~/.config/i3/config: i3/config
```

### LINKED ISSUE #97: Specify shell version for shell plugin
I use dovecot on different OS and ran into a tiny problem trying to expand curly braces like

``` yaml
[shell]
    - ["mkdir -p test/{foo,bar}", Create some directories]
```

This works on e.g. macOS but not on Ubuntu due to the differences in the `subprocess.call`'s default `sh` versions. What do you think of either using the current `$SHELL` as executable, or allowing to define a specific shell for a given command?

I would be happy to implement this if you think it's useful also for other dotbot users.

bixel:


nagromc:
I think it closes #97, not #96

bixel:
Uh, damn! Will I break anything with the PR if I rewrite the history within my fork?

nagromc:
The PR hasn't been merged yet, so I guess it's okay if you push a new history. Maybe you'll have to close this PR and open a new one, or the UI will notify us the history changed. I'm sure GitHub engineers thought about this use case ;-) Maybe a FAQ describes this case.

If you aren't confident about that, just create a dummy repository and do some tests (I'm interested in the answer too actually). I can even create a PR for you ;-)

nagromc:
Well, it seems it's [not recommend](https://help.github.com/articles/about-pull-requests/):

> When pushing commits to a pull request, don't force push. Force pushing **can** corrupt your pull request.

It **can** corrupt you PR, but they don't explain in what. Since it would be only changing a comment, and since the original forked `master` branch doesn't move fast, I think it would be seamless. But that's only my point of view.

anishathalye:
It's totally fine to force push to your own repo when updating a pull request.

plugins/shell.py
```diff
@@ -41,8 +41,10 @@ def _process_commands(self, data):
                     self._log.lowinfo(cmd)
                 else:
                     self._log.lowinfo('%s [%s]' % (msg, cmd))
+                executable = os.environ.get('SHELL') if os.environ.get('SHELL') else None
```
anishathalye: You can do `dict.get(key, default_value)` (so `os.environ.get('SHELL', None)`).

bixel: Ah ok, I'll adjust it soon, thanks.

anishathalye:
Sorry for the delay… this somehow fell off my radar.

Merged in 913d548. Thanks for the contribution!

PR #105, opened 2016-11-18
Allow empty link sources

casey:
Heyo!

Here's a diff which I'm a bit ambivalent about, but I thought I'd open a PR anyways.

It changes the link plugin to allow the source to be None. (In yaml, this means having no value, or having the value `~`.)

If the source for a link is null, it uses the basename of the destination. If the basename has a leading '.', it strips that.

So, for example, this:

```yaml
- link:
 ~/bin/git-last:           git-last
 ~/bin/backup:             backup
 ~/.zshrc:                 zshrc
 ~/.gitconfig:             gitconfig
```

Can just be:

```yaml
- link:
 ~/bin/git-last:
 ~/bin/backup:
 ~/.zshrc:
 ~/.gitconfig:
```

I'm ambivalent about this because although it does save some typing and make config files a little easier to write, it does introduce a bit of magic which is not immediately obvious.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,7 +189,9 @@ symbolic link should have a relative path.
       path: zshrc
 ```
 
-If the source location is omitted or NULL, dotbot will use the destination instead, without a leading `.` if present. This makes the following three config files equivalent:
+If the source location is omitted or set to `null`, Dotbot will use the
+basename of the destination, with a leading `.` stripped if present. This makes
+the following three config files equivalent:
 
 ```yaml
 - link:
--- a/README.md
+++ b/README.md
@@ -189,6 +189,47 @@ symbolic link should have a relative path.
       path: zshrc
 ```
 
+If the source location is omitted or NULL, dotbot will use the destination instead, without a leading `.` if present. This makes the following three config files equivalent:
+
+```yaml
+- link:
+    ~/bin/ack: ack
+    ~/.vim: vim
+    ~/.vimrc:
+      relink: true
+      path: vimrc
+    ~/.zshrc:
+      force: true
+      path: zshrc
+```
+
+```yaml
+- link:
+    ~/bin/ack:
+    ~/.vim:
+    ~/.vimrc:
+      relink: true
+    ~/.zshrc:
+      force: true
+```
+
+```json
+[
+  {
+    "link": {
+      "~/bin/ack": null,
+      "~/.vim": null,
+      "~/.vimrc": {
+        "relink": true
+      },
+      "~/.zshrc": {
+        "force": true
+      }
+    }
+  }
+]
+```
+
 ### Shell
 
 Shell commands specify shell commands to be run. Shell commands are run in the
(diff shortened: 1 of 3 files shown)


plugins/link.py
```diff
@@ -31,6 +31,12 @@ def _process_links(self, links):
                 relink = source.get('relink', relink)
                 create = source.get('create', create)
                 path = source['path']
+            elif source == None:
+                basename = os.path.basename(destination)
+                if basename[0:1] == ".":
```
anishathalye: Could you change the `"` to `'` (double quotes to single quotes)? That's the convention followed in the rest of the code.

casey:
Updated with single quotes

anishathalye:
This is a cool idea! Also, it's backwards compatible, which is awesome.

I'd like to merge this, though I'd like to wait for a week or so to see if anyone else has feedback on this before merging.

Would you mind adding some documentation to the README about this? Also, could you add a [test](https://github.com/anishathalye/dotbot/tree/master/test#testing) for this feature? (for the test, you don't need to get it set up to run the stuff locally, you could add a test and let it run on Travis CI).

casey:
Yup, that all sounds good.

I agree that waiting for feedback is a good idea.

I'll update the readme and add a test.

casey:
Is there a way to get more verbose output from the tests in travis, so I can see why the tests are failing?

anishathalye:
You can add `DEBUG=true` as the first line in the test, and that'll give you some more information.

Other than that, it's kinda messy :/

anishathalye:
Oh it also tells you which step failed. So if you look at the output, for a lot of tests, it says `run failed`. So that narrows it down to that particular line of the test.

casey:
Made it work with dict links, added tests, and mentioned it in README.

README.md
```diff
@@ -189,6 +189,47 @@ symbolic link should have a relative path.
       path: zshrc
 ```
 
+If the source location is omitted or NULL, dotbot will use the destination instead, without a leading `.` if present. This makes the following three config files equivalent:
```
anishathalye: minor: "dotbot" -> "Dotbot"
minor: hard wrap line

anishathalye: I think the "without a leading `.` present" by itself doesn't exactly specify what happens -- maybe mention that you take the basename without the leading `.` present?

plugins/link.py
```diff
@@ -46,6 +46,16 @@ def _process_links(self, links):
             self._log.error('Some links were not successfully set up')
         return success
 
+    def _default_source(self, destination, source):
+      if source == None:
```
anishathalye: minor: `==` => `is`

casey:
Addressed feedback.

Sounds good!

anishathalye:
Squashed into a single commit (d2e20c77db3811d8d0ef640a743f8dcc85075fdc) and merged (847cb101d486319da882f958b1cb58b515cc6db2). Thanks for the contribution!

casey:
Sweet! You bet :)

PR #106, opened 2016-11-19
Mention in readme that ~ is null in YAML

casey:
This is a weird YAML gotcha, and it seems worth mentioning in the readme
since dotbot users are likely to run into it.


anishathalye:
Good point.

I made a tiny change to your commit (hard wrapped the text) and merged in 295f8e31606db36c6c745dc2346e029ac2d277a2.

Thanks!

casey:
Sweet, you bet!

Issue #113, opened 2017-03-25
[Feature Request] Provide docker support for tests execution


dein0s:
Not sure what else I can write up here. Subj speaking for itself :)

dein0s:
Will provide PR today.

anishathalye:
This looks interesting, but what's the motivation for this? We already have tests running on Vagrant (which I'm pretty sure nobody uses anyways) and also on Travis CI.

dein0s:
> (which I'm pretty sure nobody uses anyways)

Exactly, that's why! :)
Okay now, jokes aside. Docker, in comparison with Vagrant, doesn't depend on VB. It's not always possible to start VB locally. Like, for example, in my case - current workstation right now already running in vm (yeah, yeah.. I know, but I still need that damn win for gaming and for my homies).
So, what's the option here for running tests locally (i work as a qa, testing already in my blood:)) and ensure that everything works fine?
Commit, wait for travis to run it, fix, squash, clean git history.. That's just awful.
And in total - why not? Script doesn't breaks anything (almost no changes made to the original files, execpt from DOTFILES var change) and allow users to have a choice.

zamber:
Docker has massive compatibility issues with host kernels, and non-existent backwards compability.

IIRC you can run vagrant under Windows and you already have VB, so just check out the repo on Win and try testing it there.

dein0s:
First of all it's just a matter of choice. As I've mentioned - this PR doesn't break anything, just adds another option for running tests in a virtual environment. Each solution has it's own pros and cons (like described [here](http://stackoverflow.com/a/16048358) for example). And both test scripts by default designed to be executed in Unix env. 

@zamber looks like you didn't understand my case (and I apologize for this, as english aint my native). My home host is under Windows, in it I run VB with Linux as a workstations for my working needs (code, test and other stuff). So, the scheme with Vagrant will look following: windows -> vb with linux -> vb for vagrant. And yes, this is possible ofc, but such scheme will consume much more resources then I want.

anishathalye:
More lines of code added are more lines of code that need to be maintained going forward, so there isn't nonzero cost.

If Docker is superior to Vagrant, we should _switch_, not have both. But I don't think Docker is strictly better than Vagrant.

zamber:
My point was that you can use Vagrant under Windows, so there's no need for
nesting VMs. Just clone the repo and run the tests after installing Vagrant
on your Windows host. Easier than nesting flaky Docker under a VBox VM.

[...]

PR #114, opened 2017-03-25
Docker support for tests execution

dein0s:
Related to #113 
Added script for running tests in docker container with appropriate python version. Modified README to reflect changes.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -1,27 +1,45 @@
 Testing
 =======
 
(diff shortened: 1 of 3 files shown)


robobenklein:
I only looked at the code for ~5 minutes, but any reason not to use `--rm` with `docker run --rm <aoeu>` instead of checking for and removing the container on new tests?

anishathalye:
Sticking with Vagrant for now (see discussion in #113).

Issue #112, opened 2017-03-21
Need some clarification on `clean` command


dein0s:
First of all I want to thank you for this project. Just found it today and in love with it :)
Also, before we continue here, it worth to mention, that english aint my native, so I appologize for any mistakes and misunderstandings in advance.

And now to the point - my question is why
> Only dead links that point to the dotfiles directory are removed.

https://github.com/anishathalye/dotbot/blob/master/plugins/clean.py#L39

Is there any funny story behind this check? Like someone unmounted folder on a prod environment, Dotbot deleted broken link and after remounting folder back - server just crushed?

Seriously, I mean, it seems totally illogicaly for me: we have a dead link, that points to nowhere, and instead of clean/overwrite, we just leave it for the gods of errors during futher installation process.
Like, for example, I've a link ```~/.zshrc_alias -> /dev/null``` that was pointing to some real object in the past, but somehow it's gone now. And... I can't delete it and therefore create a new link that will point to my new file ```~/.dotfiles/.zshrc_alias```.

Just want to understand such implementation before I'll begin using my patronus on it :D

anishathalye:
Glad you liked the project :)

And no, there's no funny story… the reasoning was that if there was some broken link there that apparently didn't have anything to do with your dotfiles, then maybe we shouldn't touch it, just to be safe.

If you specify `relink = True` (e.g. like [this](https://github.com/anishathalye/dotfiles/blob/master/.install.conf.yaml#L4)), then broken symlinks will still be _overwritten_ (so that'll solve the problem for the example you gave above).

dein0s:
Yeah, ```relink``` will work sor sure. But, suddenly, in multi-profile/multi-config project, it can't be inherited from another config file (```base.yaml``` for example), so I have to specify this in every file.

Would you accept a PR with some changes to logic in ```clean.py``` plugin? Like adding a key ```force``` for the instance?

dein0s:
Also, regarding to this plugin, it was meant that method ```self._clean``` always return ```True```?
https://github.com/anishathalye/dotbot/blob/master/plugins/clean.py#L28

And if so, then why do we need [this](https://github.com/anishathalye/dotbot/blob/master/plugins/clean.py#L24) ```else``` statemend with logging if we'll never reach that point?

p.s. Damn, I'm getting addicted to this thing. Even thinking about to write a plugin for handling metaconfigs :D

anishathalye:
Yes. And I guess that's unreachable code 😛

[...]

PR #116, opened 2017-03-26
Added `force` option to the `clean` plugin

dein0s:
Related to #112

### DIFF
--- a/plugins/clean.py
+++ b/plugins/clean.py
@@ -19,7 +19,7 @@ class Clean(dotbot.Plugin):
         success = True
         defaults = self._context.defaults().get(self._directive, {})
         force = defaults.get('force', False)
-        for target in self._dispatch_targets(targets):
+        for target in targets:
             if isinstance(targets, dict):
                 force = targets[target].get('force', force)
             success &= self._clean(target, force)
@@ -31,8 +31,8 @@ class Clean(dotbot.Plugin):
 
     def _clean(self, target, force):
         '''
-        Cleans all the broken symbolic links in target if that point to
-        a subdirectory of the base directory or forced to clean.
+        Cleans all the broken symbolic links in target if they point to
+        a subdirectory of the base directory or if forced to clean.
         '''
         if not os.path.isdir(os.path.expanduser(target)):
             self._log.debug('Ignoring nonexistent directory %s' % target)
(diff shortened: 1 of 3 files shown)


plugins/clean.py
```diff
@@ -17,31 +17,47 @@ def handle(self, directive, data):
...
         else:
             self._log.error('Some targets were not successfully cleaned')
         return success
 
-    def _clean(self, target):
+    def _clean(self, target, force):
         '''
-        Cleans all the broken symbolic links in target that point to
-        a subdirectory of the base directory.
+        Cleans all the broken symbolic links in target if that point to
```
anishathalye: The grammar is slightly wrong here. Something like this would be better: "Cleans all the broken symbolic links in target if they point to a subdirectory of the base directory or if forced to clean."

plugins/clean.py
```diff
@@ -17,31 +17,47 @@ def handle(self, directive, data):
 
     def _process_clean(self, targets):
         success = True
-        for target in targets:
-            success &= self._clean(target)
+        defaults = self._context.defaults().get(self._directive, {})
+        force = defaults.get('force', False)
+        for target in self._dispatch_targets(targets):
```
anishathalye: You can actually just replace this with `for target in targets` and that'll work fine for both dicts and lists (a for-in iteration over a dictionary iterates over its keys). And then you could get rid of the `_dispatch_targets` function.

plugins/clean.py
```diff
@@ -17,31 +17,47 @@ def handle(self, directive, data):
...
-                if self._in_directory(path, self._context.base_directory()):
-                    self._log.lowinfo('Removing invalid link %s -> %s' %
-                        (path, os.path.join(os.path.dirname(path), os.readlink(path))))
+                points_at = os.path.join(os.path.dirname(path), os.readlink(path))
+                if self._in_directory(path, self._context.base_directory()) or force:
+                    self._log.lowinfo('Removing invalid link %s -> %s' % (path, points_at))
                     os.remove(path)
+                else:
+                    self._log.lowinfo('Link %s -> %s not removed (force=%s). '\
+                                      'You can remove it manually.' % (path, points_at, force))
```
anishathalye: I think the "You can remove it manually." part is unnecessary / this is a bit more verbose than necessary, we could leave it at just `'Link %s -> %s not removed.'`

plugins/clean.py
dein0s: Came up with this solution after writing my own ```aptget``` plugin. I agree, as far as we don't modify the dictionary and accept only lists or dictionaries, ```_dispatch_targets``` method is unnecessary here.

anishathalye:
Squashed in a2a9e1f, merged in bf2a933, docs added in c29ba8d.

Thanks for the contribution!

PR #126, opened 2017-05-31
Add instructions for installing test env on Debian

nagromc:


test/README.md
```diff
@@ -8,6 +8,19 @@ machine state. The tests are deterministic, and each test is run in a virtual
...
 easily repeatable.
 
+Installing the Test environnement
+---------------------------------
+
+### Debian-based distributions
+
+- Install the requirements
+
+      # apt install ruby-dev vagrant
```
anishathalye: Use a fenced code block so you can indicate the language and get syntax highlighting?

Also, replace `apt` with `sudo apt-get`?

test/README.md
```diff
@@ -8,6 +8,19 @@ machine state. The tests are deterministic, and each test is run in a virtual
...
+
+- Install the [Sahara][sahara] Vagrant plugin
+
+      $ vagrant plugin install sahara
```
anishathalye: replace with fenced code block?

test/README.md
nagromc: 1. You mean replace

```
- Install the requirements

      # apt install ruby-dev vagrant
```

with

````no-highlight
- Install the requirements

```shell
# apt install ruby-dev vagrant
```
````

, right?

2. Well, I'm used to:
- [use `#` (over `$`)](https://en.wikipedia.org/wiki/Command-line_interface#Command_prompt) to tell the command needs to be executed as `root`:
> On many Unix and derivative systems, the prompt commonly ends in **`$` or `%` if the user is a normal user, but in `#` if the user is a superuser** ("root" in Unix terminology).

- run `apt` over `apt-get` as it seems to be the [new standard](https://www.debian.org/doc/manuals/debian-handbook/sect.apt-get.en.html#sect.apt-get) for Debian-based distributions:

> APT is a vast project, whose original plans included a graphical interface. It is based on a library which contains the core application, and `apt-get` is the first front end — command-line based — which was developed within the project. **`apt` is a second command-line based front end provided by APT which overcomes some design mistakes of `apt-get`**.

**Comment by nagromc** [comment]
Well, according to Atom's [Markdown preview plugin](https://atom.io/packages/markdown-preview), adding the language used for formatting inside a fenced code block (`` ```shell `` instead of `` ``` ``) doesn't highlight the content. Worse: it is grayed out because of the `#`, parsed as a comment:

Code:
````
```shell
# apt install ruby-dev vagrant
```
````

Rendering in Github:
```shell
# apt install ruby-dev vagrant
```

So I just let it like

````
```
# apt install ruby-dev vagrant
```
````

to make it render like

```
# apt install ruby-dev vagrant
```

It's up to you: `# apt ...` or `$ sudo apt ...`

anishathalye:
Oh cool, I didn't know the thing about `apt`.

Well, to be consistent with the main README.md, it might be nice to

    ```bash
    <just write the command execution without a $ or anything>
    ```

I prefer `sudo apt`.

nagromc:
There you go @anishathalye  :-)

anishathalye:
I don't think `shell-session` is a supported language in GFM. How about using `bash`?

anishathalye:
Sorry for the long wait. Squashed in 174949d and merged in f8f4696!

PR #130, opened 2017-07-12
Fix shell example in README (add description)

klausweiss:
I was confused when reading the docs and I had too look up in the source code whether `description` would work.


anishathalye:
Good idea. Merged!

Thanks for the contribution.

Issue #133, opened 2017-08-14
Force should noop if the files are identical


x89:
I think when forcing the recreation of files if you've said to force recreate them yet they match identically then the file shouldn't be overwritten.

This would stop edge case bugs where a config's change is picked up during the remove/creation process and the program bugs out. It would also involve less filesystem changes.

It should be quite a simple change in one location with the use of `filecmp.cmp`.

anishathalye:
Hmm, I'm actually not sure about this. I think if you have a file in your dotfiles repository `~/src/dotfiles/foo` and a **copy** of that file in `~/foo`, then it should be overwritten with a symlink. Otherwise, if you edit `~/foo`, you won't notice that if you run a `git status`, and if you subsequently run a `./install`, then your changes will be destroyed by the old copy. I think having a symbolic link all the time helps avoid this confusion.

anishathalye:
Closing because this is intentional. If there's anything new, feel free to reopen.

PR #134, opened 2017-08-14
Filecmp for not deleting, see issue #133

x89:


anishathalye:
Thanks for taking the time to submit a patch.

I think the current behavior is intentional -- we want to delete and create a symlink (see more info in #133)

Issue #141, opened 2018-01-20
[Feature] Link all in folder


robobenklein:
We can link folders with simple syntax like
```yaml
- link:
    ~/.config:
```
But let's say I wanted to link everything in `./config/*` into `~/.config/` so that it would link all the folders in `~/.config` that are in the repo already, but the million other things in `~/.config` don't show up back in the git repo.

Perhaps a new syntax style purpose for trailing slashes:
```yaml
- link:
    ~/.config/: config/*
```
Would link all the files/folders in `./config/` to their places in `~/.config/`

This would make it extremely easy to automatically link a lot of configs in the repo with just one entry in the YAML.

Taking this idea further we could get:
```yaml
- link:
    ~: home
```
Which would take all the files/folders in `./home/*` or the repo and link them each into `~/`. (`./home/` could have things like `.tmux.conf` and `.bashrc` in it which would all be linked to `~/`.)

I had this idea as I was trying to do something similar using an `ln` command like `ln -ns $(pwd)/config/* $HOME/.config/` however this command nonzero exits when any of the files exist, or with `-f` I didn't want it to overwrite without making any backup.

anishathalye:
Seems like a neat idea; would be happy to implement this / accept a PR as long as there are no backwards compatibility issues.

Are there any issues with `*`? Technically, it's legal to have it in a file name.

Also, with the second syntax you proposed:

```yaml
- link:
   ~: home
```

How would you distinguish between wanting to symlink a directory vs wanting to individually symlink the files within it?

arcticicestudio:
One idea to solve these edge cases (files named `*`, folders/files named `home`) would be to add a CLI option e.g. `--enable-globs` which must be implicitly added by the user, otherwise dotbot should handle them like "normal" files/folders.

Another way might be to add a boolean flag (default to `False`) to the `link` plugin (e.g. `use_globs`) to enable name globbing individually for each entry.

robobenklein:
Yeah I missed trailing slashes on my home example.

I like your idea of the `use_globs` flag, being explicit is always good, I'll see what I can do to implement that.

anishathalye:
The `use_globs` sounds like a good idea, we can implement in the same way like `create` and `relink` and `force`.

Still doesn't solve the problem when you don't have the trailing `*` though. Like if you're interpreting this `~: home` or even `~: home/`, currently, Dotbot will do the equivalent of symlinking `~` to point to the `home` folder in the current directory.

[...]

PR #142, opened 2018-01-27
Implement globbing support

robobenklein:
Phew! Turns out this does a lot better than I had initially planned!

I need help making the test cases, here's the behaviour I tried to create and need to test:

Failure cases
---
```yaml
link:
  ~/bin/:
    path: bin
    use_glob: true
```
```yaml
defaults:
  - link:
      use_glob: true
link:
  ~/bin/:
    path: bin/
```
Because we aren't sure if we should link `bin` inside `~/bin` or link `bin` *as* `~/bin`.
Handled with error message by the `glob_star_loc is -1 and destination[-1] is '/'` conditional.

Success Cases
---
```yaml
link:
  ~/bin:
    path: bin
    use_glob: true
```
```yaml
defaults:
  - link:
      use_glob: true
link:
  ~/bin:
    path: bin
```
Works as normal link operation `~/bin` -> `bin`, so that globbing can be turned on at default level.

```yaml
defaults:
  - link:
      use_glob: true
link:
  ~/bin/:
    path: bin/*
```
```yaml
link:
  ~/bin:
    path: bin/*
```
Links all the items in `bin/*` into individual links in `~/bin/`. So long as there is a wildcard `*` in the `path:` then it will link the glob results as individual items into the destination.

Trailing Slash
---
`~/bin` vs `~/bin/` is a special case used to enforce a certain behavior, with a trailing slash and `use_glob` enabled, it forces dotbot to treat the link task as "link items *into* this folder" instead of potentially linking *as*.

Advanced Quirks
---
```yaml
link:
  ~/.config/tmux/:
    path: tmux/**.conf
    create: true
```
Links all the `.conf` files in `tmux/` into `~/.config/tmux/` however the double star recursive match is not supported in python's default `glob`. (but is working in [glob2 pip package](https://pypi.python.org/pypi/glob2)
Example:
```
tmux
├── 2.1/
│   ├── bind-pane-zoom.conf -> ../bind-pane-zoom.conf
│   ├── keybinds-vi.conf -> ../keybinds-vi.conf
│   ├── mouse-mode-new.conf -> ../mouse-mode-new.conf
│   └── powerline-py3.conf -> ../powerline-py3.conf
├── 2.5/
│   ├── bind-pane-zoom.conf -> ../bind-pane-zoom.conf
│   ├── keybinds-vi.conf -> ../keybinds-vi.conf
│   ├── mouse-mode-new.conf -> ../mouse-mode-new.conf
│   └── powerline-py3.conf -> ../powerline-py3.conf
├── bind-pane-zoom.conf
├── keybinds-vi.conf
├── mouse-mode-new.conf
├── mouse-mode-old.conf
├── powerline-py3.conf
├── py2u.25.tmux.conf
└── tmux.conf
```
After link operation:
```
/home/robo/.config/tmux
├── bind-pane-zoom.conf -> /home/robo/code/configs/tmux/bind-pane-zoom.conf
├── keybinds-vi.conf -> /home/robo/code/configs/tmux/keybinds-vi.conf
├── mouse-mode-new.conf -> /home/robo/code/configs/tmux/mouse-mode-new.conf
├── mouse-mode-old.conf -> /home/robo/code/configs/tmux/mouse-mode-old.conf
├── powerline-py3.conf -> /home/robo/code/configs/tmux/powerline-py3.conf
├── py2u.25.tmux.conf -> /home/robo/code/configs/tmux/py2u.25.tmux.conf
└── tmux.conf -> /home/robo/code/configs/tmux/tmux.conf
```
Notice the folders were not linked. They could have been included as links by simply changing the glob link target to be `tmux/*` instead of `tmux/*.conf`.

If [glob2 was in use](https://github.com/miracle2k/python-glob2/), we could do recursive globs and make the glob match detection better.

I don't think we should bother including glob2 because I think this feature set covers all the use cases we are concerned about, and the cost of inclusion of another library is not worth the benefit here IMO.

Even without glob2, we can still do some interesting and more useful advanced things, like:
```yaml
link:
  ~/.config/:
    path: config/*/*
    create: true
```
This example creates normal directories for everything at the `~/.config/*` level, but then everything at the `~/.config/*/*` level is now a symlink.

Anyways,
---
I looked at the test setup, but it's a bit too cryptic for me to dive into tonight. Could someone else try making the test cases that match these?

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,17 +165,22 @@ files if necessary. Environment variables in paths are automatically expanded.
 
 Link commands are specified as a dictionary mapping targets to source
 locations. Source locations are specified relative to the base directory (that
-is specified when running the installer). Directory names should *not* contain
-a trailing "/" character.
+is specified when running the installer). If linking directories, *do not* include a trailing slash.
 
 Link commands support an (optional) extended configuration. In this type of
 configuration, instead of specifying source locations directly, targets are
-mapped to extended configuration dictionaries. These dictionaries map `path` to
-the source path, specify `create` as `true` if the parent directory should be
-created if necessary, specify `relink` as `true` if incorrect symbolic links
-should be automatically overwritten, specify `force` as `true` if the file or
-directory should be forcibly linked, and specify `relative` as `true` if the
-symbolic link should have a relative path.
+mapped to extended configuration dictionaries.
+
+Available extended configuration parameters:
+
+| Link Option | Explanation |
+| -- | -- |
+| `path` | The target for the symlink, the same as in the shortcut syntax (default:null, automatic (see below)) |
+| `create` | When true, create parent directories to the link as needed. (default:false) |
+| `relink` | Removes the old target if it's a symlink (default:false) |
+| `force` | Force removes the old target, file or folder, and forces a new link (default:false) |
+| `relative` | Use a relative path when creating the symlink (default:false, absolute links) |
+| `use_glob` | Treat a `*` character as a wildcard, and perform link operations on all of those matches (default:false) |
 
 #### Example
 
@@ -207,6 +212,10 @@ the following three config files equivalent:
     ~/.zshrc:
       force: true
       path: zshrc
+    ~/.config/:
+      use_glob: true
+      path: config/*
+      relink: true
 ```
 
 ```yaml
@@ -217,6 +226,10 @@ the following three config files equivalent:
       relink: true
     ~/.zshrc:
       force: true
+    ~/.config/:
+      use_glob: true
+      path: config/*
+      relink: true
 ```
 
 ```json
@@ -230,6 +243,11 @@ the following three config files equivalent:
       },
       "~/.zshrc": {
         "force": true
+      },
+      "~/.config/": {
+        "use_glob": true,
+        "path": "config/*",
+        "relink": true
       }
     }
   }
--- a/dotbot/dispatcher.py
+++ b/dotbot/dispatcher.py
@@ -30,10 +30,11 @@ class Dispatcher(object):
                         try:
                             success &= plugin.handle(action, task[action])
                             handled = True
-                        except Exception:
+                        except Exception as err:
                             self._log.error(
                                 'An error was encountered while executing action %s' %
                                 action)
+                            self._log.debug(err)
                 if not handled:
                     success = False
                     self._log.error('Action %s not handled' % action)
--- a/plugins/link.py
+++ b/plugins/link.py
@@ -1,4 +1,5 @@
 import os
+import glob
 import shutil
 import dotbot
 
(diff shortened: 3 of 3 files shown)


robobenklein:
https://github.com/anishathalye/dotbot/issues/141

robobenklein:
It seems to me that since this doesn't break any backwards compatibility it could get merged before 2.0 (https://github.com/anishathalye/dotbot/issues/35)

Although while developing I think I've found another change which I'll take a look at when it's not 4am.

anishathalye:
This looks pretty good so far! Don't worry about implementing tests -- if you describe test cases to me, I can implement them.

Btw, we should also update the documentation in the readme to explain this new globbing behavior.

robobenklein:
Hopefully the table format is easy enough to understand. When I first found dotbot it took me awhile to find all the options so perhaps that will help.

What's the separation between the Github Wiki and the Readme? Should I add or update something in the wiki if it's already been covered by the Readme?

Also, I think the fail/success cases in the first comment should be good to go with the current code. Let me know if anything else needs to be changed.

kevinrambaud:
Hey guys, I'm was browsing the repo looking for this exact feature to integrate/use. Great job btw, is there anything I can do to speed up the integration of this feature?

anishathalye:
Thank you for your patience. And thanks again for the patch, I really like the feature!

I made one small change to the API: I renamed `use_glob` to just `glob` (ae7ef1e).

I got around to implementing tests -- check out the `glob` branch, specifically e1a7863. It seems that there is a bug in the current implementation. In particular, the pattern `~/bin: bin/*` does not work correctly -- it links files to e.g. `~/bina`, `~/binb`, etc. rather than `~/bin/a`, `~/bin/b`, etc. Doing `~/bin/: bin/*` works properly, though. I think the issue is that we shouldn't be manipulating paths with string manipulation operations, in `glob_link_destination = destination + glob_item`. Instead, this should be an `os.path.join`. I fixed this in 94c58e5.

Let me know how the `glob` branch looks. I think it's ready to merge, but I'll give you some time to give any additional input in case you have anything to add, and default to merging in 1 week otherwise.

kevinrambaud:
Great job @anishathalye I'm going to play with it and keep you guys informed

anishathalye:
Also just FYI the tests are slightly broken on Travis CI but I'll fix that in a couple min hopefully.

anishathalye:
Ok fixed in ee3d208.

anishathalye:
Merged in 2f4cc0d9cb70235e7c5820dc4e7dd89b1d375fb5!

Thanks again for the contribution.

dsifford:
@anishathalye Can you clarify this: https://github.com/anishathalye/dotbot/issues/149#issuecomment-381296887

anishathalye:
Nope, `clean` does not accept globs at the moment. If you have a design or an implementation that makes sense, we can adopt it.

dsifford:
Would it be wise to implement something that adds dangerous symlinking without the ability to clear the symlinks?

No disrespect, but it seems like this PR was merged too quickly. Applications that create "default" config files, for example, will see dead symlinks in the config locations and not create default thinking that these dead links are actual config files. This, I'd imagine, could run the risk of breaking functionality of applications, no?

anishathalye:
Dead symlinks can always be cleared manually. And this new functionality doesn't affect existing configuration files, only ones that use the new `glob: true` flag.

In general, when migrating between different configurations, it will always be possible to do it in such a way that'll leave dead symlinks lying around: this was already a possibility before this patch.

Perhaps a better version of `clean` will do a recursive directory traversal and remove all dead symlinks that point inside the dotfiles directory.

dsifford:
Correct, but without the built-in ability of recursive clean, never before has there been a possibility of having a `link` strategy that wasn't also cleanable.

Globbing introduces that.

anishathalye:
I'm having a hard time seeing how glob makes things worse: I think this was already a problem. Consider an example based on the one from #149:

```yaml
- link:
    ~/.config/tmux/:
      path: tmux/*/*.conf
      glob: true
```

which produces:

```
tmux
├── 1.9
│   ├── keybinds-vi.conf -> ../keybinds-vi.conf
│   ├── mouse-mode-old.conf -> ../mouse-mode-old.conf
│   └── powerline-py3-old.conf -> ../powerline-py3-old.conf
```

Without globbing, how would I have implemented this? Perhaps like this:

```
- link:
    ~/.config/tmux/1.9/keybinds-vi.conf: tmux/1.9/keybinds-vi.conf
    ~/.config/tmux/1.9/mouse-mode-old.conf: tmux/1.9/mouse-mode-old.conf
    ~/.config/tmux/1.9/powerline-py3-old.conf: tmux/1.9/powerline-py3-old.conf
```

If I, say, renamed `1.9` to `2.0`, then all the symlinks in `~/.config/tmux/1.9/*` would be broken. I could add a line `clean: ~/.config/tmux/1.9` at that point, but that solves the problem both before and after the glob plugin.

dsifford:
Right, you could add in the non-glob in clean, but that creates a disparity between your link and clean actions. 

In my mind, and perhaps this is only me, I prefer to be deliberate about both links and cleans. If I chose to link AND clean `~/.config/tmux/1.9`, then renaming the `1.9` folder would cause the install step to fail and I'd be forced to adjust, deliberately, my configuration to reflect the new structure.

I'll stick with merge since it's working for me. Seems like you're okay with functionality as it currently sits, which is fine as well.

Issue #146, opened 2018-03-08
[Question] Link rather than coping install script shim


DEVoytas:
Are there any reasons why not to do this:

```
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ If you're using Git, you can add Dotbot as a submodule:
 cd ~/.dotfiles # replace with the path to your dotfiles
 git init # initialize repository if needed
 git submodule add https://github.com/anishathalye/dotbot
-cp dotbot/tools/git-submodule/install .
+ln -s dotbot/tools/git-submodule/install install
 touch install.conf.yaml
```
Benefits I see:
- avoiding missing changes when install shim is updated in upstream
- avoiding file duplication

Are there any downsides I am missing?

anishathalye:
Yeah - when you clone your dotfiles repo for the first time, the symlink won't work (because you will not have fetched the dotbot submodule yet). You could address this with `git clone --recursive` but I think it's better to copy the shim. Also, the shim hasn't been updated since October 2014, and it's probably unlikely to change any time soon.

DEVoytas:
Hi @anishathalye 
Thanks for a really quick response!
OK, I see your point. On the other hand you will not be copy it either (until you fetch the submodule), and having a dead link at least gives some info on what you are missing.
Anyway, I already created an PR: https://github.com/anishathalye/dotbot/pull/147 (before I read your response), but feel free to decline it, if you don't like it.

anishathalye:
I think the issue is with multiple machines. For example, I have my dotfiles cloned on tens of machines; I wouldn't want to remember to use `--recursive` every time I clone on a new machine.

DEVoytas:
Thanks for clarification @anishathalye 
Closing it.

PR #147, opened 2018-03-08
README: Link rather than coping install script shim (address #146)

DEVoytas:
Here is the PR, in case it turns out to be not such a bad idea.


DEVoytas:
Closed because it does not make sense as explained in https://github.com/anishathalye/dotbot/issues/146

PR #150, opened 2018-04-14
[WIP] LinkIf shell conditionals

robobenklein:
Still need to do:

- [ ] Check Python 2/3 compatibility
- [ ] Update docs
- [ ] Describe behaviour clearly
- [ ] Decide how verbose to be with the command stdout and stderr

Example yanked from my own configs: (using ZSH as my shell, commands are evaluated there)
```yaml
- linkif:
  # Conky
    ~/.conky/unseptium:
      if: 'conky --version && [[ $(hostname) == "robo-unseptium" ]]'
      create: false
      path: conky/unseptium
      force: true
  # TF2
    ~/unhexium/SteamLibrary/steamapps/common/Team Fortress 2/tf/cfg/:
      if: '[[ -d ~/unhexium/SteamLibrary/steamapps/common/Team\ Fortress\ 2 ]]'
      create: false # don't want this on non-gaming systems
      path: tf2/*
      glob: true
      force: true
```

Note carefully the use of single and double quotes, and escaping the spaces in paths. I'm not sure if we want to change anything about quoting or escaping levels, but it seems to work well for me.

### DIFF
--- a/dotbot/dispatcher.py
+++ b/dotbot/dispatcher.py
@@ -34,7 +34,7 @@ class Dispatcher(object):
                             self._log.error(
                                 'An error was encountered while executing action %s' %
                                 action)
-                            self._log.debug(err)
+                            self._log.debug("Err: " + str(type(err)) + ": " + str(err))
                 if not handled:
                     success = False
                     self._log.error('Action %s not handled' % action)
--- a/dotbot/dispatcher.py
+++ b/dotbot/dispatcher.py
@@ -43,6 +43,7 @@ class Dispatcher(object):
     def _load_plugins(self):
         self._plugins = [plugin(self._context)
             for plugin in Plugin.__subclasses__()]
+        self._log.debug("Found plugins: " + str(self._plugins))
 
 class DispatchError(Exception):
     pass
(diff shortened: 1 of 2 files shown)


anishathalye:
Why a separate linkif rather than integrating this functionality into the standard link. Or better yet, into all the built-in plugins?

robobenklein:
Sure, I can easily move an `if`-check into the main plugin, I guess I don't really know what behaviour other people want yet.

How should I implement the `if` conditional into the other plugins besides link? The same way?

anishathalye:
I'm not sure what you mean by main plugin. You shouldn't modify the `Plugin` abstract base class.

Hmm, on second thought, I guess it doesn't really make sense to implement `if` for `clean`. And even for `shell`, it seems unnecessary (I could be convinced either way on that one).

robobenklein:
Note for creating tests: ensure that at least some form of the test covers the failure fixed by that latest commit:

```
testing conditional: conky --version && [[ $(hostname) == "robo-unseptium" ]]
Command returned 127: 
conky --version && [[ $(hostname) == "robo-unseptium" ]]
An error was encountered while executing action link
Err: <class 'RuntimeError'>: dictionary changed size during iteration
Action link not handled
```

PR #155, opened 2018-05-21
Add --no-realpath-base switch to keep symbolic links in BASEDIR

ypid:
Note: This is one of two competing PRs which try to solve the same problem. It is open for discussion which one of those should be merged (or none of them). Ref: https://github.com/anishathalye/dotbot/pull/156

This PR changes the behavior for all plugins when the `--no-realpath-base` switch is used. It has not been carefully tested what this effects. It is beleaved that it does not break stuff because normally the OS resolves symlinks for you when you open a file. Only when deleting, there might be differences in practice which should not occur for the base directory.

dotbot had a hardcoded behaviour that the BASEDIR was always passed to
os.path.realpath which "returns the canonical path of the specified
filename, eliminating any symbolic links encountered in the path".

This might not always be desirable so this commit makes it configurable
using a command line flag, `--no-realpath-base`.

The use case where `--no-realpath-base` comes in handy is the following:
You want to provide dotfiles in the Filesystem Hierarchy Standard under
`/usr/local/share/ypid_dotfiles/`. Now you want to provide
`.config/dotfiles` as a default in `/etc/skel`. When you now
pre-configure `/etc/skel` by running dotbot in it set has HOME, dotfiles
will refer to `/usr/local/share/ypid_dotfiles/` and not
`/etc/skel/.config/dotfiles` which does not look nice.

This is related to but not the same as the `relative` parameter used
with link commands.

### DIFF
--- a/dotbot/cli.py
+++ b/dotbot/cli.py
@@ -17,6 +17,8 @@ def add_options(parser):
     parser.add_argument('-d', '--base-directory', nargs=1,
         dest='base_directory', help='execute commands from within BASEDIR',
         metavar='BASEDIR', required=True)
+    parser.add_argument('-r', '--no-realpath-base', action='store_false',
+        dest='base_directory_real_path', help='Do not eliminate symbolic links encountered in BASEDIR')
     parser.add_argument('-c', '--config-file', nargs=1, dest='config_file',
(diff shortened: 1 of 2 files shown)


anishathalye:
I'm in agreement with you, in that I prefer #156 over #155, so I'm going to close this PR (and we can continue discussion in #156).

PR #156, opened 2018-05-21
Add `real_path` option to link command allow keeping source with symlink

ypid:
Note: This is one of two competing PRs which try to solve the same problem. It is open for discussion which one of those should be merged (or none of them). Ref: https://github.com/anishathalye/dotbot/pull/155. They where written in the order they where opened as PR. @ypid prefers this (#156) variant, because it allows greater flexibility and has no impact on any plugins because it only effects the link command.

dotbot had a hardcoded behaviour that the BASEDIR was always passed to
os.path.realpath which "returns the canonical path of the specified
filename, eliminating any symbolic links encountered in the path".

This might not always be desirable so this commit makes it configurable.

The use case where `real_path` comes in handy is the following:
You want to provide dotfiles in the Filesystem Hierarchy Standard under
`/usr/local/share/ypid_dotfiles/`. Now you want to provide
`.config/dotfiles` as a default in `/etc/skel`. When you now
pre-configure `/etc/skel` by running dotbot in it set has HOME, dotfiles
will refer to `/usr/local/share/ypid_dotfiles/` and not
`/etc/skel/.config/dotfiles` which does not look nice.

This is related to but not the same as the `relative` parameter used
with link commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,7 @@ Available extended configuration parameters:
 | `relink` | Removes the old target if it's a symlink (default:false) |
 | `force` | Force removes the old target, file or folder, and forces a new link (default:false) |
 | `relative` | Use a relative path to the source when creating the symlink (default:false, absolute links) |
+| `canonicalize-path` | Resolve any symbolic links encountered in the source to symlink to the canonical path (default:true, real paths) |
 | `glob` | Treat a `*` character as a wildcard, and perform link operations on all of those matches (default:false) |
 | `if` | Execute this in your `$SHELL` and only link if it is successful. |
 | `ignore-missing` | Do not fail if the source is missing and create the link anyway (default:false) |
--- a/dotbot/context.py
+++ b/dotbot/context.py
@@ -1,4 +1,5 @@
 import copy
+import os
 
 class Context(object):
     '''
@@ -13,8 +14,11 @@ class Context(object):
     def set_base_directory(self, base_directory):
         self._base_directory = base_directory
 
-    def base_directory(self):
-        return self._base_directory
+    def base_directory(self, canonical_path=True):
+        base_directory = self._base_directory
+        if canonical_path:
+            base_directory = os.path.realpath(base_directory)
+        return base_directory
 
     def set_defaults(self, defaults):
         self._defaults = defaults
--- a/dotbot/dispatcher.py
+++ b/dotbot/dispatcher.py
@@ -10,8 +10,8 @@ class Dispatcher(object):
         self._load_plugins()
 
     def _setup_context(self, base_directory):
-        path = os.path.abspath(os.path.realpath(
-            os.path.expanduser(base_directory)))
+        path = os.path.abspath(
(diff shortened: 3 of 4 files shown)


anishathalye:
Hi, I must have lost track of this, just re-reading it now. I agree with you -- I prefer this PR #156 over the alternative #155 that you proposed. My reasons are: (1) this functionality only changes the behavior of link, not all plugins; the current CLI options are only for global things, and (2) I think not that many people use Dotbot's CLI options -- it's nicer to have everything in the config file, in my opinion.

This seems like a reasonable feature to include. I'm not super happy with the name -- having an option called `real_path` seems a little confusing. I wonder if we could come up with a more self-documenting name. (I haven't come up with anything better so far.) Thoughts?

Also, a minor thing -- even if we keep `real_path`, maybe we can rename it to `real-path` to match the style of other options (like the recently added `ignore-missing`).

ypid:
Thanks for picking this PR up.

Another name could be canonical path. Lets maybe go with canonical path then? I guess "real_path" is a term that is only familiar to developers (that is at least my impression when I search for both of these terms online).

anishathalye:
What do you think about changing it from noun to verb/imperative form, e.g. `canonicalize-path`? That will match `create` / `relink` / `force` / `ignore-missing`.

ypid:
Works for me :)

anishathalye:
I squashed/rebased your commits here: https://github.com/anishathalye/dotbot/commit/8bced45c7de3f17f5cbfc02d5122f3caacd0d52b

I made a small fix that I think ends up being useful for using `canonicalize-path` practically: https://github.com/anishathalye/dotbot/commit/8bced45c7de3f17f5cbfc02d5122f3caacd0d52b#diff-ab8fc2f6853cb329ab55a90d95657968L76-R79

I added a couple tests just to make sure my understanding of `canonicalize-path` is correct. Do you think these tests make sense: https://github.com/anishathalye/dotbot/commit/6572677a71a69ee825bd851841dab0dccc5489d5? In short, I'm creating the following `dotfiles` and `dotfiles-symlink -> dotfiles`, a symlink to the dotfiles directory. When Dotbot is called within `dotfiles-symlink` and run with the defaults, it should resolve it and any symlink created should have `dotfiles` in it, e.g. `.foo -> /home/user/dotfiles/foo`. But when `canonicalize-path: false` is set, then it should be `.foo -> /home/user/dotfiles-symlink/foo`.

ypid:
Looks good to me :+1: I also tested your branch and it works for me. Also thanks for writing the integration tests kinda for me! :)

One note, my git commits are GPG signed. Such signatures are destroyed by squah/rebase. I now did a GPG signed squash. If you want, you can cherry-pick your changes ontop of this commit and then merge it into master. Just if you are interested :)

anishathalye:
Great, thanks, I'll use your commit that has the signature intact. Merged in ec8498ffb88d43c7d0648033d3d758ebb96a9378. Thank you for the pull request!

PR #157, opened 2018-05-24
Update lib/pyyaml submodule with the current `pyyaml` repository

mfeliu:
The current `pyyaml` repository is in GitHub: https://github.com/yaml/pyyaml


anishathalye:
Thanks for pointing this out! There are some more complexities that need to be addressed here, including modifying the install shim. Even then, the old anishathalye/pyyaml repository needs to be updated because people might be using the old shim with a new Dotbot version. Commit 76884e9a0c5e79de1c27361d196b6aba3c87b214 implements these changes (and it also upgrades the PyYAML dependency version). Let me know what you think.

If there are no new comments, I'll merge 76884e9a0c5e79de1c27361d196b6aba3c87b214 in a couple days.

PR #159, opened 2018-05-24
Fix handling of base directory

anishathalye:
follow-up from discussion at https://github.com/anishathalye/dotbot/commit/76884e9a0c5e79de1c27361d196b6aba3c87b214

cc @robobenklein

### DIFF
--- a/dotbot/cli.py
+++ b/dotbot/cli.py
@@ -14,10 +14,10 @@ def add_options(parser):
         help='suppress most output')
     parser.add_argument('-v', '--verbose', dest='verbose', action='store_true',
         help='enable verbose output')
-    parser.add_argument('-d', '--base-directory', nargs=1,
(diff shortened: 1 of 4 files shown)


robobenklein:
Well this fix was a lot easier than I was expecting.

I was assuming we were going to just fix globbing because link already had references and handling for the basedir / context_dir. (But if this really is it - they could be simplified now)

I don't suppose there's any reason not to just chdir, right? Am I missing anything?

anishathalye:
Yeah, I don't think there's a reason to not chdir. It was effectively happening before with the shim+dotbot, but it makes more sense to chdir in dotbot because people may use a different shim or not use it at all.

Currently not passing the tests because I'm dumb and typed the wrong thing, but I'll fix that in a minute.

PR #160, opened 2018-05-30
Add PyPI package

### LINKED ISSUE #89: Create a python package
It would be useful to have a package to manage dotfiles outside ~/, something like this (considering you already have a dotfiles repo installed and configured):

```
pip install dotbot
mkdir project
cd project
git init
dotbot --install gitignore-language-template
```

Where the file is inside the dotfiles dir and the install command would copy instead of symlinking. Where is the dotfiles dir? I guess that could be configured inside a dotfile ;)

This would make the dotfiles templating management works for projects outside ~/, which is an nice-to-have feature

anishathalye:
This patch also makes the '-d' argument optional, with the base
directory defaulting to the directory of the configuration file.

Fixes #89 

cc @gilmrjc @dsifford @paride

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,4 @@
+*.egg-info
 *.pyc
+build/
+dist/
--- a/README.md
+++ b/README.md
@@ -74,6 +74,17 @@ submodule; be sure to commit your changes before running `./install`, otherwise
 the old version of Dotbot will be checked out by the install script. If using a
 subrepo, run `git fetch && git checkout origin/master` in the Dotbot directory.
 
+If you prefer, you can install Dotbot from [PyPI] and call it as a command-line
+program:
+
+```bash
+pip install dotbot
+touch install.conf.yaml
+```
+
+In this case, rather than running `./install`, you can invoke Dotbot with
+`dotbot -c <path to configuration file>`.
+
 ### Full Example
 
 Here's an example of a complete configuration.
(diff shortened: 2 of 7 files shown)


dsifford:
anishathalye is the man!

dsifford:
Getting this error when attempting to use the new package:

```
(ins) $ dotbot --help
Traceback (most recent call last):
  File "/home/dsifford/.local/bin/dotbot", line 7, in <module>
    from dotbot import main
  File "/home/dsifford/.local/lib/python3.6/site-packages/dotbot/__init__.py", line 1, in <module>
    from .cli import main
  File "/home/dsifford/.local/lib/python3.6/site-packages/dotbot/cli.py", line 4, in <module>
    from .config import ConfigReader, ReadingError
  File "/home/dsifford/.local/lib/python3.6/site-packages/dotbot/config.py", line 4, in <module>
    from .util import string
ModuleNotFoundError: No module named 'dotbot.util'
```

Looks like the "util" module isn't included in the package...

```
dsifford ~/.../python3.6/site-packages/dotbot
(ins) $ tree
.
├── cli.py
├── config.py
├── context.py
├── dispatcher.py
├── __init__.py
├── plugin.py
└── __pycache__
    ├── cli.cpython-36.pyc
    ├── config.cpython-36.pyc
    ├── context.cpython-36.pyc
    ├── dispatcher.cpython-36.pyc
    ├── __init__.cpython-36.pyc
    └── plugin.cpython-36.pyc

1 directory, 12 files
```

anishathalye:
Welp. Thanks for pointing this out -- I tested with `pip install -e .` but I should have checked this too...

It looks like all the directories under `dotbot/*` (`messenger` and `util`) are missing. Investigating now (any advice welcome too -- I'm not super familiar with setuptools).

dsifford:
Hah no problem. I'm probably much less familiar with it than you, but happy to help if I can.

PR #162, opened 2018-06-02
Fix missing packages in PyPI distribution

### LINKED ISSUE #161: PyPI package is broken
See https://github.com/anishathalye/dotbot/pull/160#issuecomment-394047536

cc @dsifford

anishathalye:
Fixes #161


anishathalye:
I think this fixes it. I tested by copying the `.whl` file to another machine and installing from that.

Will merge tomorrow if there are no comments between now and then.

PR #168, opened 2018-08-06
fix clean path expansion to also expand environment variables

dsifford:
Pretty self-explanatory here.

The salient use-case of this fix is the expansion of XDG environment variables.

So, for example, these three configurations would be equivalent (if using default XDG locations):

```yml
- clean:
    - '~'
    - ~/.local/bin
    - ~/.config
```

```yml
- clean:
    - $HOME
    - $HOME/.local/bin
    - $XDG_CONFIG_HOME
```

```yml
- clean:
    - /home/dsifford
    - /home/dsifford/.local/bin
    - /home/dsifford/.config
```

Let me know if I missed something. Thanks!

### DIFF
--- a/dotbot/plugins/clean.py
+++ b/dotbot/plugins/clean.py
@@ -34,11 +34,11 @@ class Clean(dotbot.Plugin):
         Cleans all the broken symbolic links in target if they point to
         a subdirectory of the base directory or if forced to clean.
         '''
-        if not os.path.isdir(os.path.expanduser(target)):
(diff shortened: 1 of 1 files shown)


anishathalye:
Good catch! I'll add a test for this and merge tonight.

anishathalye:
Merged cf55ca51881f4364774fde03fb5446f027cb0d74 (implementation) and 9332cde5ad5c7ab27483b63f61f1a8de6836743a (tests) in f197ededb11e7613aa0f4c5f88311e9ab4653749 (tagged v1.12.5).

Thanks again for fixing this.

PR #169, opened 2018-09-20
Feature: Conditional Links

robobenklein:
Linking with the `if` option.

Has been stable for me for months now on multiple systems with even *very* outdated python installs.


robobenklein:
My examples:

```
  # Conky
    ~/.conky/unseptium:
      if: 'conky --version && [[ $(hostname) == "robo-unseptium" ]]'
      create: false
      path: conky/unseptium
      force: true

  # TF2
    ~/unhexium/SteamLibrary/steamapps/common/Team Fortress 2/tf/cfg/:
      if: '[[ -d ~/unhexium/SteamLibrary/steamapps/common/Team\ Fortress\ 2 ]]'
      create: false # don't want this created on any other systems
      path: tf2/*
      glob: true
      force: true
```

anishathalye:
Thank you for the contribution!

I squashed your commits and slightly simplified the implementation in e45cb6fdaa4b919f76f1041be2df06f3fba81ee5. Let me know what you think.

I'll write tests for this tomorrow, and then it should be ready to merge.

robobenklein:
Am I missing something or does it look like the `if` in https://github.com/anishathalye/dotbot/compare/linkif#diff-efa7fceeeb8e708aebdb684de044b03dR36 doesn't actually stop the linking?

But I completely agree with that approach, wish I had thought to move it deeper into the call stack.

anishathalye:
Merged in 1b17e8e.

PR #171, opened 2018-10-06
Add option --no-color to suppress colorization of output

phallstrom:
By default, if output is a TTY, dotbot will colorize the output. This
patch adds the option to pass `--no-color` to dotbot to have it suppress
this colorization.

My terminal colors are such that the ones used by dotbot aren't useful and in some cases are hard to read (see attached image).

I tried to add tests, but unfortunately all test output is piped so no colors would appear anyway.

I tried running tests through `script` which tricks python into thinking is is a TTY, but ran into problems  somewhere between vagrant and the test setup that I couldn't figure out.

no-color-before-and-after


anishathalye:
Seems fine to not have tests for this.

This feature is a great idea, thanks for the pull request! Merged.

Issue #170, opened 2018-09-29
Shell - Some commands were not successfully executed


shivapoudel:
I am receiving Some commands were not successfully executed message with this below config :) 
![capture](https://user-images.githubusercontent.com/3774827/46244483-1a4b0100-c3ff-11e8-9bd3-d8662fa48650.PNG)

This is my `install.conf.yaml` file. I am on Windows 10 with Dotbot 1.12.5:
```
- defaults:
    link:
      force: true
      create: true
      relink: true

- clean: ['~']

- link:
    ~/.gitconfig: git/gitconfig
    ~/.gitignore_global: git/gitignore_global
    ~/.minttyrc: mintty/minttyrc
    ~/.npmrc: npm/npmrc
    ~/.vim: vim/vim
    ~/.vimrc: vim/vimrc

- shell:
  - vim +PluginInstall +qall
```

anishathalye:
Looks like your `vim +PluginInstall +qall` failed. Some steps I'd take when debugging:

- try running it directly in your terminal to see if it returns successfully
- run it with `stdout: true` / `stderr: true` in Dotbot and see if it prints any errors

shivapoudel:
Modify config shell setup to like this:
```
- shell:
    -
      command: vim +PluginInstall +qall
      stdout: true
```
No luck @anishathalye same error and I don't get any output too

anishathalye:
Did you try running it directly? Also, did you try enabling stderr too?

shivapoudel:
@anishathalye I am using Mintty which is shipped in Git for Windows. Tried stderr too but no luck :)

anishathalye:
What happens if you run the command directly?

shivapoudel:
It work with direct execution :)

![capture](https://user-images.githubusercontent.com/3774827/46261313-a6e5e400-c511-11e8-9d59-d11cefc3ca40.PNG)
![capture-1](https://user-images.githubusercontent.com/3774827/46261314-a6e5e400-c511-11e8-9a54-5ec3a3735843.PNG)

anishathalye:
I don't use vim plug, so I'm not very familiar with it. Does it require input on stdin before proceeding? Also, this issue seems somewhat related: https://github.com/junegunn/vim-plug/issues/225

shivapoudel:
@anishathalye solve through running Git CMD with administritive privilage instead of Git Bash.

anishathalye:
Nice, glad you got it working!

PR #177, opened 2019-02-04
don't set executable arg to subprocess.call on Windows

SuJiKiNen:
as https://bugs.python.org/issue2304 said,

executable given

`C:\\Program Files\\Git\\usr\\bin\\bash.exe`

this PR is a workaround for windows.

Possiable relate to #170 

if you have to expand command like '~', you
should straight tell to use git bash like
`"bash -c 'your command'"` ,make sure that bash
is in your  `PATH`.

### DIFF
--- a/dotbot/plugins/link.py
+++ b/dotbot/plugins/link.py
@@ -24,6 +24,8 @@ class Link(dotbot.Plugin):
         success = True
         defaults = self._context.defaults().get('link', {})
         for destination, source in links.items():
+            destination = os.path.normpath(destination)
+            source = os.path.normpath(source)
             destination = os.path.expandvars(destination)
             relative = defaults.get('relative', False)
             force = defaults.get('force', False)
--- a/dotbot/plugins/link.py
+++ b/dotbot/plugins/link.py
@@ -24,9 +24,9 @@ class Link(dotbot.Plugin):
         success = True
         defaults = self._context.defaults().get('link', {})
         for destination, source in links.items():
+            destination = os.path.expandvars(destination)
             destination = os.path.normpath(destination)
             source = os.path.normpath(source)
-            destination = os.path.expandvars(destination)
             relative = defaults.get('relative', False)
             force = defaults.get('force', False)
             relink = defaults.get('relink', False)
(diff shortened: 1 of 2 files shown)


anishathalye:
The discussion in the link makes it seem like the issue has been fixed?

SuJiKiNen:
yeah,I dig a little deeper into the source.
[https://github.com/python/cpython/blob/4a66524006852fc982aebafa277f2c043d9ad149/Lib/subprocess.py#L977](url)
if we pass executable,
the right way to call is ` bash -c "cmd.exe /c ls"`
but actually we call  is `bash  "cmd.exe /c ls"`
and it will throw  `bash: cmd.exe /c ls: No such file or directory`
we don't have `-c` option.
that's why I suggest do not pass executable arg,
and users should be responsible to their shell commands,like `bash -c [command]`
another fix could be we just do some check for executable,and add  `-c ` option
for the executable when under Windows with git bash.
what do you think?

mohkale:
Just to throw my two sense here. I wouldn't classify the discussion on #170 as a fix. More of a workaround. It seemed to me like dotbots shell directive wasn't working with git bash on windows, so they used cmd.exe instead. I had the same issue, with shell directives just outputting `/c: /c: Is a directory` (while using git bash on windows).

In my case the issue was two fold, first 'git bash on windows' doesn't seem to export the SHELL environment variable, which meant dotbot was defaulting to cmd.exe as the shell. I fixed it by modifying `./install` to `export SHELL`.

The second issue is that bash fails to start (with the aforementioned error message). I fixed it by removing the `shell=True` line from `dotbot/plugins/shell.py` and modifying the command to be `[executable, '-c', cmd]`. This isn't a cross platform fix, if you want to install on windows you have to install a bash like shell (or at least one that takes the `-c` flag), so I understand if you won't want to make this change in the actual repo.

mohkale:
adding on to the previous point, I feel like issues like this would be a lot less prominent if dotbot had an API for executing shell commands. Something as simple as a `run_process.py` file which had a a method called `run_process`. As it stands different plugins appear to be using the same general format of `shell.py` meaning, at least in regards to this issue, we have to keep monkey patching the methods to make them work.

anishathalye:
Thank you for reporting this issue and providing a pull request. I used your proposed strategy as the core of #220.

Issue #180, opened 2019-03-17
Create missing target files/folders from source


qwerbzuio:
I recently migrated from another dotfile-manager, where there was an option called `add` to move an existing file to the dotfile-repository and create a link to the moved file. I really liked that option, as it eases inclusion of new dotfiles. I was a bit confused that `dotbot` doesn't seem to provide something similar (I think that the creator of #135 had the same issue).

I think that `dotbot` could easily implement such a feature as default behaviour. If the target is missing and the source exists, it could simply move source to target and then create a link from the source location. From my point of view, this is the only sensible thing `dotbot` can do in that situation (except from doing nothing, which is the current behaviour since the fix for #103).

I've prepared a pull request with a possible implementation for the behaviour described in this issue.

anishathalye:
Thanks for the feature suggestion / PR.

I think this does not fit the Dotbot model: it's nice to have things be explicit rather than implicit. Plus, it's a one-time "setup cost" to move the file into the dotfiles repository, so I think this doesn't save that much work.

WilliamYuhangLee:
> I think this does not fit the Dotbot model: it's nice to have things be explicit rather than implicit.

Then is it possible to add a new command called "move" or "add" so we can automate this process explicitly? Though it is indeed a one-time cost but I feel like such a feature is essential for a dotfile manager and it is easy to implement anyway.

anishathalye:
Hmm, but then the command in the `.install.conf.yaml` would only be useful for the

ewen-lbh:
I don't understand how this can be a "one-time" cost: when I'll eventually install new software, it'll create its config file in ~/.config, and I'll need to:

- move it over
- add the 'link' entry
- symlink it back

not mentioning that adding a 'link' entry could also be abstracted away to a simple command on `dotbot/bin/dotbot`, eg `dotbot link SOURCE [TARGET]` where `TARGET` would default to simply `SOURCE`, so that the `.dotfiles` dir is a simple mirror of the home directory

anishathalye:
Closing because I'm not planning on adding this feature at this time.

PR #181, opened 2019-03-17
Create missing target files/folders from source

qwerbzuio:
If the target in the dotbot-repo doesn't exist, copy from the original. The links are then created afterwards.

This implements the behaviour described in #180.


anishathalye:
Thank you for taking the time to submit a PR. As described in #180, I'd rather keep this explicit than make Dotbot do it automatically.

PR #185, opened 2019-06-04
Fix `path` parameter's docs for linking feature

ronalabraham:
- small update to the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,11 +160,11 @@ Available extended configuration parameters:
 
 | Link Option | Explanation |
 | -- | -- |
-| `path` | The target for the symlink, the same as in the shortcut syntax (default:null, automatic (see below)) |
+| `path` | The source for the symlink, the same as in the shortcut syntax (default:null, automatic (see below)) |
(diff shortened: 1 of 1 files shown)


ronalabraham:
Also, should [this warning](https://github.com/anishathalye/dotbot/blob/v1.15.0/dotbot/plugins/link.py#L87) say "source" instead of "target". I'm assuming the terminology is such that target -> source (and destination -> source), consistent with `ln`'s man page.

anishathalye:
I agree that our wording is currently confusing. Your changes to the README.md look good in terms of making the wording within the README itself consistent.

As you point out, `link.py`'s error messages are also a bit confusing in terms of the source/target wording: the error messages use "target" to refer to "what the symbolic link points to", in the sense of being a "link target". But yeah, `ln` uses "source" to refer to that concept.

Shall we change the wording in the error messages at the same time? (if you grep for "target" in `link.py`, you'll find the 3 uses of that word)

ronalabraham:
@anishathalye Done. I've always found `ln`'s terminology counterintuitive but makes sense to be consistent.

PR #188, opened 2019-06-25
Fix - Shell scripts to check python binary in windows

### LINKED ISSUE #187: [Bug] Don't find python library from WindowsApps dir
I accidentally installed the Python library from Windows App which was the culprit for me. I am receiving this error message:

```
/c/Users/shivapoudel/.dotfiles/meta/dotbot/bin/dotbot: line 10: /c/Users/shivapoudel/AppData/Local/Microsoft/WindowsApps/python3: Permission denied
/c/Users/shivapoudel/.dotfiles/meta/dotbot/bin/dotbot: line 10: /c/Users/shivapoudel/AppData/Local/Microsoft/WindowsApps/python3: Permission denied
```

Later, I debug and found https://github.com/anishathalye/dotbot/blob/master/bin/dotbot#L10-L12 check is locating me there. Can this check be modified so that `which python3` is located from other place rather than `C:\Users\shivapoudel\AppData\Local\Microsoft\WindowsApps`

shivapoudel:
This closes #187 which I am facing repeatedly due to accident installation of python via the Windows App store. Windows App store always adds it if I sign in with Windows account and no rules over there to delete it permanently. Thus making PR over here.

### DIFF
--- a/bin/dotbot
+++ b/bin/dotbot
@@ -7,9 +7,9 @@
 # is useful because we don't know the name of the python binary.
 
 ''':' # begin python string; this line is interpreted by the shell as `:`
-which python3 >/dev/null 2>&1 && exec python3 "$0" "$@"
 which python  >/dev/null 2>&1 && exec python  "$0" "$@"
 which python2 >/dev/null 2>&1 && exec python2 "$0" "$@"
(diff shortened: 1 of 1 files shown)


anishathalye:
Something similar done in 9a8d292681ac540894dce3fe998be0068dfa2df5, see discussion in #187.

Issue #90, opened 2016-07-12
make a back up of overwiritten file or folder


QiangF:
Hello everyone,
When the force option is true, dotbot removes the original file. That could be risky. Will it be better if the original file or folder is renamed with a version suffix ~1~, ~2~, like emacs did for its backup files?
Thank you!

petersandor:
I think this was requested before, see #12

anishathalye:
I don't think there's any need to change the existing behavior for force. After all, it is called "force" 😛.

What would you think about having a separate `backup: true` option for this kind of behavior.

I didn't really like it back in #12, but I think enough people have been asking for this that we can add it.

QiangF:
The use cases are:
1 when you want to try another set of dotfiles (tens of files) and you don't have enough time to go through each of them, have a backup makes you worry free. 
2 when you dist-upgrade to a new version of the os, some configuration may change, but most of them may still work,  you can try the old dotfiles and let the problem to appear by themselves, only if you have a backup.

anishathalye:
Ok, let's make this happen. I like having a `backup: true` option, which specifies that files should be backed up and then removed and relinked.

Does anyone want to submit a PR for this?

ibeex:
hi I added backup option it is backup: dir_name dir_name is name off sub directory inside dotfile dir in that directory backup off existing files is created. Also I added one more option. Adding non existing entries in config file, they will be added and recreated with link. code is here https://github.com/ibeex/dotbot

ashyon:
I don't think the backup location should have to be inside the dotfiles directory - it has nothing to do with your current dotfiles and probably shouldn't be synced to git

ibeex:
@ashyon For most scenarios you are probably right. I was using principle better safe then sorry. I don't think there is to much overhead if backup dir is in dotfiles directory. If you don't need it in git, it can always go in .gitignore.

c-c-k:
Hello, having a limited background in Python, wanting to become an open source contributor and needing to better organize my dot files I came upon Dotbot and am hoping that I can be of some use.

While looking through the source code, issues and pull requests, this issue caught my attention as I constantly worry that I might erase something important by mistake.

For the moment I started by updating the code presented by @ibeex  [a few comments above](https://github.com/anishathalye/dotbot/issues/90#issuecomment-279323350) to the current upstream code base.

In regards to the questions presented in [PR 190](https://github.com/anishathalye/dotbot/pull/190#issuecomment-522343238):

[...]

Issue #36, opened 2015-05-04
Replace with backup


warang580:
As I used Dotbot to create my dotfiles repository, I have seen other tools that allows to replace current "real" files by the new symlink and putting them in a `backup` folder instead of doing nothing is there is already a file. That would be really useful, as I had to it myself manually.

It could be something like :

`$ ./install --replace`

If you implement it, in the `dotfiles_template` repository you could add a `backup` folder with its content gitignored.

anishathalye:
This was discussed before in #12.

I think this might be a good thing to include in Dotbot 2.0. I don't think it's a huge pain point for now, because it's a one-time thing when deploying to a new system. For updates, it shouldn't come up.

I agree that it's a nice feature to have, and perhaps default options (#34) might be the best way to implement this. You could have something like:

```yaml
default:
  link:
    backup: ./backup
    replace: true
```

Issue #12, opened 2015-02-09
Backup


airondumael:
It will be good if there is an option that will make dotbot backup the existing files and write over it.

anishathalye:
Thanks for the suggestion!

Hmm, where should the backup file be backed up, `${file}.bak`? What if that file exists already? We could back it up to `${file}.bak.${number}`, but then that gets a bit ugly.

This seems to be a one-time thing per machine, and it should probably require user interaction (to merge information from old config files, etc) or at least alert the user. I'm not sure what's the best way of designing this feature. Maybe it's best to leave it up to the user, the way it is now?

anishathalye:
I think this is best left to the user.

I'm going to go ahead and close it for now. Feel free to reopen it if you have any ideas.

lliseil:
Voting for such a feature.
Existing dotfiles could be backed up in ~/.dotfiles-backup/, or ~/.dotfiles-backup/timestamp/ Allowing us to quickly set up our dotfiles, knowing we can easily check, merge and revert if needed.

anishathalye:
Hmm, I guess we can reopen this issue.

I still don't know what the best way of handling this is. Would every *file* go to the backup folder? What if there are name collisions? Should the backup folder mirror the directory hierarchy instead?

I think there's some stuff to figure out before this could be implemented.

And as of now, I still think it's best left to the user -- in the majority of cases that I can think of, this isn't something that needs to be handled multiple times, and so it's probably handled best on a case by case basis anyways.

lliseil:
A KISS proposal
* simply backup any replaced (aka deleted) dotfiles; eventually dropin directory hierarchy
* local, ie in $HOME out of the `dotfiles` dir.
* a `backup-timestamp` folder would avoid any filename collisions I can think of (unless server's a real mess :) Since it's a local folder

That only would add serenity, when eg in a hurry to deploy.

anishathalye:
Let's leave this as-is for now.
I don't know if the hurry-to-deploy scenario happens often enough to warrant the additional complexity in the code.

PR #190, opened 2019-08-16
Add backup option

JonasDe:
This pull request introduces a 'backup' option to prevent loss of data that could occur with the 'force' option in the link directive.


anishathalye:
Thanks for the pull request. It might be worth reading through discussion of backup in previous issues/PRs such as #12, #36, #90 so we can decide on a design.

How exactly do you imagine this being used? Where will the backup directory be? (in the dotfiles directory?) Will it be specified per-file, or globally? What happens if there's a conflict?

anishathalye:
Closing due to inactivity; feel free to re-open if anything new comes up.

PR #192, opened 2019-09-15
Update README to clone submodule 

### LINKED ISSUE #191: Track master branch in submodule and ignore dirty commits
Hi,
So i found out that when dotbot is added as a submodule, it automatically detaches itself from the head, practically having its own branch. After much research i found that providing ` -b <branch>` automatically tracks the submodule brach, making it `git submodule add -b master https://github.com/anishathalye/dotbot`.

Another issue I was having, was that whenever i ran dotbot it automatically created a *dirty* commit in the dotbot submodule. This is also preventable by adding ` ignore = dirty ` in the `.gitmodules` file under respective submodule. 

This solves issues when trying to automate dotfiles provisioning with ansible, in my case. 

Could the README be updated with this information?

darsh12:
This resolves #191

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,10 +39,11 @@ If you're using **Git**, you can add Dotbot as a submodule:
 ```bash
 cd ~/.dotfiles # replace with the path to your dotfiles
 git init # initialize repository if needed
-git submodule add https://github.com/anishathalye/dotbot
+git submodule add -b master https://github.com/anishathalye/dotbot # Track master branch of dotbot, instead of detaching head
 cp dotbot/tools/git-submodule/install .
 touch install.conf.yaml
 ```
+Edit `.gitmodules` and add `ignore = dirty` under the dotbot submodule, to prevent tracking *dirty* commits.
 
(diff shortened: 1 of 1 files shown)


anishathalye:
What exactly does explicitly tracking `master` buy us? I think the semantics of all the `git submodule` subcommands are such that if no branch is specified, the commands behave as if `branch = master`.

Re: the discussion in #191, I think I don't understand exactly what you are saying. Submodules are always "detached" in the sense that they are just a checkout of a particular commit (and I think this is the behavior we want here: cloning your dotfiles should use the version of Dotbot you added or upgraded to, not the latest version on `master`).

darsh12:
> cloning your dotfiles should use the version of Dotbot you added or upgraded to, not the latest version on master

I did not think about it like that. The reason i wanted to track the master branch was that ansible would not clone the submodule if the submodule did not sync with the master branch.

darsh12:
Since the functionality of the submodule is supposed to be detached, i am closing this pull request

PR #194, opened 2019-09-16
Ignore dirty commits

darsh12:
As discussed in #191


anishathalye:
Thanks! Merged 76082e1 in f6c83b8.

Issue #199, opened 2019-10-04
`create` directive?


jesseleite:
I love the simplicity of how symlinking works, but is it possible to add a `create` directive, which would create a directory (and any parent directories, if necessary), _without symlinking_?  Right now I'm running this as a shell command:

```
  - description: Creating vim undo cache
    command: "[ -d ~/.vim/undo ] || mkdir ~/.vim/undo"
```

anishathalye:
Sure, this seems reasonable. What should the format be?

```
- mkdir:
    - ~/.vim/undo
    - ~/path/to/some/other/folder
```

jesseleite:
I would say:

```
- create:
    - ~/.vim/undo
    - ~/path/to/some/other/folder
    - ~/path/to/some/other/blankfile.txt
```

With `create` instead of `mkdir`, because you already abstract terminology like `shell` and `link`, opposed to `bash` and `ln`.  Also, I'm not sure when creating a blank file might be desirable, but why not support it as well?

jesseleite:
Also not sure how far I'll get, but going to attempt a PR for Hacktoberfest 😅 ...if that's okay with you?

anishathalye:
Sure, PRs are always appreciated 😄 

Agree re: abstract terminology.

Hmm, I don't know if we can handle both files and folders (also, are empty files useful?). Consider your example (repeated below for convenience):

```
- create:
    - ~/.vim/undo
    - ~/path/to/some/other/folder
    - ~/path/to/some/other/blankfile.txt
```

Here, how would Dotbot know that `folder` is supposed to be a folder, rather than a file with no file extension? Similarly, how would it know that `blankfile.txt` is a file, rather than a folder that happens to have a "." in the name?

jesseleite:
I think files could be useful for creating blank log files, though I don't have a concrete example.  I figure if `link` can link both files and folders, why not `create` both files and folders?

That said, I think you're right about detecting file vs folder, and files should be an explicitly optional thing...

```
- create:
  - path: ~/.vim/undo
    recreate: true
  - path: ~/path/to/some/other/blank-log-file.txt
    recreate: true
    file: true
```

What are your thoughts on that?

anishathalye:
What are the semantics of `recreate`?

jesseleite:
Basically like `relink` and `force`, but for the new `create` directive.  Would remove the old target and re-create it.

anishathalye:
[...]

PR #200, opened 2019-10-06
`create` directive

jesseleite:
Implementing a `create` directive (see issue #199 for details). Example...

```yaml
- create:
    - ~/path/to/some/cache/folder
    - ~/path/to/some/blank-log-file.txt
```

- [x] Create empty folders.
- ~~[ ] Create empty files.~~ (Not going to address in this PR)

### DIFF
--- a/dotbot/dispatcher.py
+++ b/dotbot/dispatcher.py
@@ -27,8 +27,8 @@ class Dispatcher(object):
                     # keep going, let other plugins handle this if they want
                 for plugin in self._plugins:
                     if plugin.can_handle(action):
-                        success &= plugin.handle(action, task[action])
                         try:
+                            success &= plugin.handle(action, task[action])
                             handled = True
                         except Exception as err:
                             self._log.error(
--- a/dotbot/dispatcher.py
+++ b/dotbot/dispatcher.py
@@ -27,8 +27,8 @@ class Dispatcher(object):
                     # keep going, let other plugins handle this if they want
                 for plugin in self._plugins:
                     if plugin.can_handle(action):
+                        success &= plugin.handle(action, task[action])
                         try:
-                            success &= plugin.handle(action, task[action])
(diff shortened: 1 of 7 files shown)


dotbot/dispatcher.py
```diff
@@ -27,8 +27,8 @@ def dispatch(self, tasks):
                     # keep going, let other plugins handle this if they want
                 for plugin in self._plugins:
                     if plugin.can_handle(action):
+                        success &= plugin.handle(action, task[action])
                         try:
-                            success &= plugin.handle(action, task[action])
```
jesseleite: This is temporary while I'm debugging, didn't mean to commit this sorry.

anishathalye:
Ooh nice, you've even written a test 😄 

See the discussion in #199. Perhaps we should only have creating directories, and skip the file creation part?

jesseleite:
Btw, how about I clean this PR up for folder creation, and work on file creation in a separate PR?  It's kinda a different discussion, and that way you can critique it separately ☺️

anishathalye:
Yes, that sounds good! I think folder creation is a good idea, but file creation might require more discussion.

jesseleite:
Alright, think I'm done with this PR.  Take a look when you get a chance, no rush though ☺️

anishathalye:
Thanks! From a quick skim, this looks good. I appreciate the attention to detail, like the change to the `shell` example. Thanks for the unit test as well!

I will try to take a closer look, and then squash and merge, in the next day or two.

jesseleite:
Cool man, sounds good thank you ☺️

anishathalye:
Squashed in 5a0f6676d41a7291e89f5e32e09396490d71a800 and merged in 04c113b5b85e87969fdbbab22d26bc4aa5398e69. Thanks again for your contribution!

Issue #203, opened 2019-11-09
use `command -v` instead of `which`


paulohefagundes:
I'd like to ask if you could replace `which` for `command -v`.

I use Arch Linux and recently Arch no longer installs `which` by default through the [base package](https://www.archlinux.org/news/base-group-replaced-by-mandatory-base-package-manual-intervention-required/). The maintainers explained some reasons why `command -v` is better here
- https://www.reddit.com/r/archlinux/comments/de1er6/arch_linux_news_base_group_replaced_by_mandatory/f2v8uhu/

Currently as a workaround I installed `which` locally, but pretty much just because dotbot uses it to find python. I'd like to be able to remove `which` from my system once dotbot no longer depends on it

anishathalye:
Thank you for the contribution! I hadn't realized that `which` isn't a builtin while `command` is (and is POSIX) -- I learned something today 😄 

Fixed in #204.

PR #204, opened 2019-11-09
replace `which` for `command -v`

paulohefagundes:
addresses #203


anishathalye:
Thanks for the contribution! Rebased in d2913e6 (edited commit message to include info from https://github.com/anishathalye/dotbot/issues/203) and merged in 2dc876c.

PR #214, opened 2020-03-26
Add --only and --except command-line arguments

### LINKED ISSUE #44: CLI flag to only run certain sections?

anishathalye:
Internal to Dotbot, we use the name "skip" instead of "except", because
the latter is a keyword, and using a name like "except_" didn't seem as
nice.

Fixes #44

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ Dotbot makes installing your dotfiles as easy as `git clone $url && cd dotfiles
 - [Configuration](#configuration)
 - [Directives](#directives) ([Link](#link), [Create](#create), [Shell](#shell), [Clean](#clean), [Defaults](#defaults))
 - [Plugins](#plugins)
+- [Command-line Arguments](#command-line-arguments)
 - [Wiki][wiki]
 
 ---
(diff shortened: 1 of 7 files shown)


anishathalye:
Replying to https://github.com/anishathalye/dotbot/issues/44#issuecomment-604491525 here (cc @mohkale).

Yep, set the metavar.

Oh interesting, I hadn't thought about Dispatcher becoming part of the public API. I suppose we could stick this in the context. Is the plugin open-source, so I could take a look at it?

mohkale:
Not open source (more just a haphazard put together mess of a plugin I use to make hierarchical and composable configs). If you want to see it, it's [here](https://github.com/mohkale/.dotfiles/blob/master/setup/plugins/subbot.py).  It depends on a few other scripts in the same directory, so if you want to try it out you'll also need [run_process](https://github.com/mohkale/.dotfiles/blob/master/setup/plugins/run_process.py) and [log_mixin](https://github.com/mohkale/.dotfiles/blob/master/setup/plugins/log_mixin.py).

mohkale:
ping.

anishathalye:
This seems like an interesting example of Hyrum's law -- I did not expect plugins to be instantiating their own dispatchers. I hadn't really thought of Dispatcher as part of the public API. Similarly, it feels a little strange to expose `only` / `except` in the context available to plugins.

For now, I will merge this PR as-is (I'm happy with `--only` / `--except` becoming part of the command-line interface). If at a later date we decide to expose those to plugins, that can be done without changing the command-line interface.

mohkale:
Well, Hyrums law is new to me. Cool 😎.

Issue #128, opened 2017-06-15
Add copy config option


jerrykal:
Add an option to copy file instead of symlink, e.g.
```yaml
- link:
    ~/.zshrc:
      copy: true
      path: zshrc
```

anishathalye:
Would it be better to just have a separate `copy:` thing like `link:` ?

ooJerryLeeoo:
That's a good idea.

sharethewisdom:
Also, options could be added to `copy:` equivalent to options found in `install(1)`?

anishathalye:
Actually, it may be better to have a `copy: true` option (as @ooJerryLeeoo suggested) rather than a separate `copy:` task (otherwise, there would be a lot of duplicated logic).

anishathalye:
As discussed in #221, we couldn't find concrete cases where `copy:` was necessary. This issue also hasn't received a ton of attention, so perhaps this feature is not needed.

obionex-5:
Hi! I think I have a use case for that feature. I want to track my standard Dockerfile and if dotbot links it then docker-compose can't use this symlink as a file. Because docker-compose is strict about working context.
It would be great to be able to instruct dotbot to copy files instead of linking them in some specific cases.

anishathalye:
I don't use docker-compose, but based on [this issue](https://github.com/docker/compose/issues/4979) it seems like it does support symlinks (except on Windows)?

obionex-5:
I'm on a mac. Symlinks does not seem to work for a Dockerfile.

anishathalye:
Ah, I see. For now, you can use `shell:` with a command like `cp -u ...`.

jmigual:
Hi I found another use case. I have a `.gitconfig` file that I use both for Windows and Linux. It wouldn't be much of a problem except when signing commits with GPG where Git on Windows requires you to define the gpg.program key if you don't want it to use the incorporated gpg from Git-Bash. So, my idea was to comment the Windows only part of the `.gitconfig` file, copy the file and uncomment the part if I am on Windows.

eengstrom:
@jmigual - I offer an alternative for your use case that I use:

First, I define an environment variable, `${DBOT_TARGET_DISTRO}`, which is set in my`dotbot` wrapper script.  Then, with a file `~/.gitconfig-windows` or `~/.gitconfig-linux` or whatever, you can link the `distro-local` file like this:

```
- link:
    # add some distro-specific symlinks
[...]

PR #221, opened 2020-05-08
Implement simple copy plugin

walkingice:
to address #128 , this PR implements a simple copy plugin. It only recognizes 3 options

* create
* force
* skippable

this plugin was rewriting from link plugin, so some functions are very similar with functions of link plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,34 @@ the following config files equivalent:
       relink: true
 ```
 
+### Copy
+
+Copy command copies files or directories to destination.
+
+#### Format
+
+| Copy Option | Explanation |
+| -- | -- |
+| `path` | The source file to copy, the same as in the shortcut syntax (default:null, automatic (see below)) |
+| `create` | When true, create parent directories to the destination as needed. (default:false) |
+| `force` | Force removes the old target, file or folder, and forces a new copy (default:false) |
+| `skippable` | If old target exists, skip this copy (default:true) |
+
+#### Example
+
+```yaml
+- copy:
+    ~/.config/terminator:
+      create: true
+      path: config/terminator
+    ~/.bash_history:
+      path: bash_history
+      skippable: true
+    ~/.vimrc:
+      path: vimrc
+      force: true
+```
+
 ### Create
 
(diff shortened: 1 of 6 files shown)


walkingice:
I think `link` and `copy` are in different context, so I created a new plugin, instead of adding option `copy:true` to `link`.

In most of cases, `link` is good enough. But I found in my use cases, sometimes I still need `copy`, so I reuse `link` plugin to create this simple `copy` plugin.

walkingice:
Thanks for your response!

I agree DSL complexity should be reduced. just my opinion: `link` plugin reminds user the linux command `ln`, adding option `copy: true` is not easy to understand, because there is not any concept about copying inside `ln`. 

in terms of examples, I think `copy` is like providing default config to a new system. In my use case, I have 2 laptops, one for work, one for personal. Both laptops need similar `~/.gitconfig`, but with different username and email. So I only keep the common part of gitconfig in my repo, then `copy` file to different laptop.

I was thinking to refactor `link` first, but I found refactoring it is not easy. So I chose implementing `copy` first. Once we know which part of code should be shared, we could refactor it again.

anishathalye:
I don't fully understand the gitconfig use case / how `copy` is required here (are your dotfiles open-source? I could just take a look at the code if that's easier).

I have a similar situation where my machines share a common `~/.gitconfig`, but some details differ. The way I handle it is that my main dotfiles symlink `~/.gitconfig`, and I use [this configuration](https://github.com/anishathalye/dotfiles/blob/master/gitconfig#L113-L115) to include `~/.gitconfig_local`. This file is also symlinked into place; I have a separate repo with per-machine configuration (in branches). For example, [here](https://github.com/anishathalye/dotfiles-local/blob/mac/gitconfig_local) is the `~/.gitconfig_local` I use on my laptop (`mac` branch). For all of this, I use `link`, not `copy`.

walkingice:
thanks for the trick, I don't know that until now. :D

it seems `copy` is not necessary in any of your cases, just let me close this PR.

Thanks for your reply and this great project!

Issue #81, opened 2016-03-30
Named groups of tasks


kazmasaurus:
Something I'm currently doing is grouping related commands by their goal rather than their type, i.e. instead of having all the links together, I have:

```yaml
#### vim ####
- shell:
    - [git submodule update --init --recursive "vim/vim-plug", Installing vim-plug]
- link:
    ~/.vimrc: vim/.vimrc
    ~/.vim/autoload/plug.vim: vim/vim-plug/plug.vim

### fish ###
- link:
    ~/.config/fish/config.fish: fish/config.fish
```

While this works great, it leads to the output of dotbot not being as easy to parse:
- It's pretty hard to see where one goal ends and the next begins, even with the green success messages.
- The success messages aren't a great fit for repetition, it's a little weird to see "All commands have been executed" multiple times.

Given this, it would be cool to lift this sort of jobs pattern into dotbot itself, so that dotbot can intelligently group output based on the jobs. I'm thinking something like this:

```yaml
- jobs:
    vim: 
        - shell:
            - [git submodule update --init --recursive "vim/vim-plug", Installing vim-plug]
        - link:
            ~/.vimrc: vim/.vimrc
            ~/.vim/autoload/plug.vim: vim/vim-plug/plug.vim
    fish:
        - link:
            ~/.config/fish/config.fish: fish/config.fish
```

(I'm not particularly for or against the top level `jobs` task, but it seems like the easiest way to fit into the current plugin capability.)

anishathalye:
Hmm, interesting idea. Can it be easily implemented in a clean way?

clintval:
I am also interested in this feature! Just started using `dotbot` last week and love it. Definitely replicates what I have been hand-scripting for months.

Even an entry like `label` which would render a horizontal labeled rule with another color would help me distinguish groupings of tasks. Over holiday here I'll see if I can sketch out a prototype.

ghalestrilo:
I really want to work on this! I had a slightly different idea, though: using an `include` keyword, which calls dotbot on other files. This would allow setup reuse and the implementation is really straightforward.

```yaml
# main.conf.yaml
- include:
  - ./vim.conf.yaml
  - ./fish.conf.yaml
```

```yaml
# vim.conf.yaml: 
- shell:
  - [git submodule update --init --recursive "vim/vim-plug", Installing vim-plug]
- link:
  ~/.vimrc: vim/.vimrc
[...]

PR #227, opened 2020-07-09
Add handling of groups

schmidtandreas:
This is the implementation of groups handler, that is discussed in #81.

### DIFF
--- a/dotbot/cli.py
+++ b/dotbot/cli.py
@@ -77,7 +77,10 @@ def main():
         if tasks is None:
             log.warning('Configuration file is empty, no work to do')
             tasks = []
-        if not isinstance(tasks, list):
+        if isinstance(tasks, dict):
+          if not 'groups' in tasks:
+            raise ReadingError('Configuration file must contains the "groups" structure')
+        elif not isinstance(tasks, list):
             raise ReadingError('Configuration file must be a list of tasks')
         if options.base_directory:
             base_directory = os.path.abspath(options.base_directory)
--- a/dotbot/cli.py
+++ b/dotbot/cli.py
@@ -41,20 +41,6 @@ def read_config(config_file):
     reader = ConfigReader(config_file)
     return reader.get_config()
 
-def precheck_tasks(log, tasks):
-    if isinstance(tasks, dict):
-        if not 'groups' in tasks:
-            raise ReadingError(
-                'Configuration file must contains the structure with the name "groups"')
-        groups = tasks['groups']
-        if not isinstance(groups, dict):
-            raise ReadingError('The groups structure has to have group names')
-        for group in groups:
-            if not isinstance(groups[group], list):
-                raise ReadingError('The group %s mast contains a list of tasks' % group)
-    elif not isinstance(tasks, list):
-        raise ReadingError('Configuration file must be a list of tasks')
-
 def main():
     log = Messenger()
     try:
@@ -91,7 +77,12 @@ def main():
         if tasks is None:
             log.warning('Configuration file is empty, no work to do')
             tasks = []
-        precheck_tasks(log, tasks)
+        if isinstance(tasks, dict):
+            for group in tasks:
+                if not isinstance(tasks[group], list):
+                    raise ReadingError('The group %s mast contains a list of tasks' % group)
+        elif not isinstance(tasks, list):
+            raise ReadingError('Configuration file must be a list of tasks')
         if options.base_directory:
             base_directory = os.path.abspath(options.base_directory)
         else:
--- a/dotbot/cli.py
+++ b/dotbot/cli.py
@@ -41,6 +41,20 @@ def read_config(config_file):
     reader = ConfigReader(config_file)
     return reader.get_config()
 
+def precheck_tasks(log, tasks):
+    if isinstance(tasks, dict):
+        if not 'groups' in tasks:
+            raise ReadingError(
+                'Configuration file must contains the structure with the name "groups"')
+        groups = tasks['groups']
+        if not isinstance(groups, dict):
+            raise ReadingError('The groups structure has to have group names')
+        for group in groups:
+            if not isinstance(groups[group], list):
+                raise ReadingError('The group %s mast contains a list of tasks' % group)
+    elif not isinstance(tasks, list):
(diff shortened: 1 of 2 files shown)


schmidtandreas:
This is the first try to implement groups handling. Currently the documentation is missing.

anishathalye:
Some high-level thoughts:

When using groups, it seems redundant to specify `groups:` as the dictionary key at the top level. Right now, Dotbot requires that the config file is a list of tasks. So I think that means that we can implement this functionality in a backwards-compatible way by assigning an interpretation to the top-level being a dictionary, and have it be the group names. So this is basically what you have, except omitting the outer "groups:". Was there a particular reason you had it at the top level?

The overloading of `self._only` might warrant a bit more thought. Right now, it's used to run a subset of plugins (e.g. just `link` and not `shell`). Making it also select groups seems a bit tricky. E.g. what would happen if someone used "link" as a group name? It's probably unlikely that someone would do that, but it seems worth thinking about. Should we add a new CLI flag for selecting groups? (e.g. `--groups group [...]`) Or is it better to keep the CLI flag set simpler?

schmidtandreas:
_Was there a particular reason you had it at the top level?_
The reason was that I assumed that use only the list / dict to differentiate between groups and tasks could prove inadequate in the future. I think I was too careful and will change it according to your suggestion.

schmidtandreas:
_Should we add a new CLI flag for selecting groups? (e.g. --groups group [...]) Or is it better to keep the CLI flag set simpler?_
That a good point. If we add two new flags: `--group-only group [...]` and `--group-except group [...]`. IMO it would be simple and understandable. A combination with the `--only task [...]` and / or `--except task [...]` would then also be possible and would enable better selectability.

anishathalye:
Do you think both only and except for group are necessary? For the plugins I think it's nice to sometimes use except (e.g. to disable the shell plugin, which can sometimes call slow scripts). I'm not sure exactly how people will end up using groups. Would just `--group` to select a subset of groups be enough?

schmidtandreas:
_Do you think both only and except for group are necessary?_
Yes, I need the except of group/-s for my tests. My use-case in detail is as follows:

I use Arch-Linux as distribution. Arch-Linux has floating release and the maintainer release every month a new version of installation image. Me and my friend using a installation script [arch-install](https://gitlab.com/schmidtandreas/arch-install) that installs, configs and setup the Arch-Linux corresponding my configuration file.

One step of this script is download, setup and installs the dotfiles using dotbot. Nearly all my dotfiles are public accessible on my [dotfile-repo](https://gitlab.com/schmidtandreas/dotfiles). Except the key-files and credentials-files, they are in a separate git-repo. This private-files-repo is only reachable on my LAN at home, or on the work LAN for the key-files for work-environment. While real installation all necessary repos and thus files are available and reachable.

To determine if something has been changed in Arch-Linux that breaks the installation or setup of the environment, we run the installation script for each configuration file on a VM once a month and check the result. This VM has no access to my private repos, neither on the home network nor on the network at work.

My solution, to avoid false positive while test is to divide the dotfiles in two groups `main` and `keys`. While real installation both groups will be processed, but while running the tests I would add `--group-except keys` to `install` execution and except the installation of the unreachable key-files. So I would continue to test the installation of most dotfiles and would not get an error due to the inaccessible key files.

That is my use-case for `--group-except` :) To be honest, I joined conversation #81 to cover exactly this use case.

schmidtandreas:
Yes, that's right, you could also cover my use case. I didn't know it, thanks for the hint! Hmm, then I'm just asking if the groups are still necessary at all. Because the profile files contain nothing but groups. The implementation of the groups would only take it in core, which is nice but not necessary.

schmidtandreas:
Maybe you could vote on it? Or you decide that this feature is not essentially necessary. I think it would be an acceptable decision because there is a solution to Issue #81 in terms of groups.

schmidtandreas:
> Maybe you could vote on it? Or you decide that this feature is not essentially necessary. I think it would be an acceptable decision because there is a solution to Issue #81 in terms of groups.

schmidtandreas:
Sorry, to close the discussion was not my intention.

schmidtandreas:
I implemented `--groups` for process only specified groups.

In addition, I tested both variants for my use case. In my view, both solutions are equivalent.
The implementation in `dotbot` of course offers the comfort that you don't have to change anything in the` install` script for the use case I have described above. However, I am not sure how many of the dotbot users actually have this or a similar use case. We would therefore add complexity into dotbot that is actually not necessary for most. This has the disadvantages that the dotbot source code becomes a little more complex and thus the maintainability becomes more difficult.
The implementation in `install` covers the use case just as well. The disadvantage is that every user has to implement and maintain it themselves.
I would prefer the `script` solution for my use case. However, there may be other use cases where the implementation in `dotbot` is better. So I stick to my statement beforehand: "I rely on your decision as maintainer" . 
If you decide to take over, I would expand the documentation accordingly. If not, let me know and I'll close this PR.

anishathalye:
Thank you for taking the time to iterate on this. It's a difficult decision, but for now, I think it's better to stick with the `script` solution.

I thought of another complexity related to this being implemented in Dotbot: once the top-level becomes a dictionary rather than an array, ordering is no longer guaranteed. Dotbot could end up executing the groups of commands "out of order" with respect to how they appear in the file, which could confuse users. We could work around this with a slightly hacky approach of having PyYAML read into an OrderedDict, but that doesn't seem ideal; we'd need an analogous hack for JSON, and having a nonstandard interpretation of YAML/JSON (with dictionaries being ordered) seems non-ideal. Seems like it's a limitation of implementing a DSL in YAML, that as it gets more complicated, things can get messy/confusing...

As I mentioned over in another issue (https://github.com/anishathalye/dotbot/issues/233#issuecomment-669596520), I'm going to try to spend some time thinking over all the Dotbot issues, see how people use Dotbot in the wild, and see if we can come up with some clean solutions to these problems without having to worry about backwards compatibility. I'll make sure to think about groups of tasks while doing that; and this iteration and discussion has definitely been helpful to my understanding of the issue, so thank you for that :)

schmidtandreas:
Thank you for your time, it also helped me a lot to understand dotbot better and to use it more effectively.

PR #231, opened 2020-07-23
Added support for python -m dotbot

TimPansino:
This tiny PR adds support for invoking dotbot using the following.

```bash
python -m dotbot
```

This is particularly useful when Python is located on your PATH, but the scripts directory of python is not.


anishathalye:
Seems useful, thanks for the PR! Released as `1.17.1` so it's available on PyPI now.

PR #245, opened 2020-11-21
link.py: globbing keep leading '.'

### LINKED ISSUE #244: Linking with glob strips leading '.' from link name
## Enviornment

OS: Ubuntu Mate 20.04.1
Python version: 3.8.2
dotbot version: 1.17.1 (yaml: 5.3)

## Setup

Given a repository with this structure:

```txt
home/.bashrc
home/.profile
home/.vim
install
install.conf.yaml
```
with this `install.conf.yaml`:

```yaml
- defaults:
    link:
        create: true
        force: true
        relink: true

- link:
    ~/:
        path: home/.*
        glob: true
```

## Execution

running `./install -v` produces this output:

```txt
Globbing with path: home/.*
Globs from 'home/.*': ['home/.vimrc', 'home/.profile', 'home/.bashrc']
Creating link ~/vimrc -> /home/erik/.dotfiles/home/.vimrc
Creating link ~/profile -> /home/erik/.dotfiles/home/.profile
Creating link ~/bashrc -> /home/erik/.dotfiles/home/.bashrc
All links have been set up

```

Running `ls -l ~` confirms that the link names do not have the leading '.'. I
would expect the link name to keep the leading '.' (e.g. `~/.vimrc`). I realize
that I can specify the links for these files individually, but this approach
is more work to maintain.

## Remarks 

If I change `path: home/.*` to be `path: home/*` based on the example from the
documentation, python glob module ignores files that start with '.'. Looking
more at how link.py works, it seems the problem might [behere](https://github.com/anishathalye/dotbot/blob/master/dotbot/plugins/link.py#L78-L80),
specifically how `glob_item` is set. I haven't put the code through a debugger
yet to confirm, but I think adding a check if `glob_base` has trailing '.',
then remove trailing '.'. This would cause `glob_item` to have the leading '.'.
I'll have some time tomorrow to try this.

etkeys:
- Updated plugins/link.py so that globbing will not strip leading '.' when
    creating link names.
- Corrected test collection output in test/link-glob-multi-start

Fixes #244

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,13 @@ Implicit sources:
       relink: true
 ```
 
+#### Limitations of `glob`
+
+Dotbot uses [glob()](https://docs.python.org/3.9/library/glob.html) to resolve
+glob paths. However, due to its design, using a glob path such as `config/*` for
+example, will not match items that being with `.`. To specifically capture
+items that being with `.`, you will need to use a path like this: `config/.*`.
+
 ### Create
 
 Create commands specify empty directories to be created.  This can be useful
--- a/dotbot/plugins/link.py
+++ b/dotbot/plugins/link.py
@@ -76,6 +76,8 @@ class Link(dotbot.Plugin):
                 else:
                     self._log.lowinfo("Globs from '" + path + "': " + str(glob_results))
                     glob_base = path[:glob_star_loc]
+                    if glob_base.endswith('/.'):
+                        glob_base = path[:glob_star_loc - 1]
                     for glob_full_item in glob_results:
                         glob_item = glob_full_item[len(glob_base):]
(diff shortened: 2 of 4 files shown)


anishathalye:
Thank you for the PR! I made a small tweak to support the glob `.*` instead of requiring `./.*`:

```patch
                 else:
                     self._log.lowinfo("Globs from '" + path + "': " + str(glob_results))
                     glob_base = path[:glob_star_loc]
-                    if glob_base.endswith('/.'):
+                    if glob_base.endswith('/.') or glob_base == '.':
                         glob_base = path[:glob_star_loc - 1]
                     for glob_full_item in glob_results:
                         glob_item = glob_full_item[len(glob_base):]
```

Rebased/squashed, merged in de8793d.

PR #248, opened 2020-11-22
Add exclude parameter for link globbing

### LINKED ISSUE #247: Add ability to exclude glob matches
Suppose I have the following repository

```txt
config/autostart/a.desktop
config/autostart/b.destkop
config/autostart/c.desktop
config/Awesome_App/settings.conf
config/Code/User/snippets/snippet1.snippet
config/Code/User/keybindings.json
config/Code/User/settings.json
config/i3/config
config/Random_App/settings.ini
install
install.conf.yaml
```

with the following `install.conf.yaml` file:

```yaml
- defaults:
    - link:
        create: true
        force: true
        relink: true

- clean:
    ~/:
        force: true
    ~/.config:
        recursive: true

- link:
    ~/.config/:
        path: config/*/*
        glob: true
```

After execution I will have symlinks to all the files specified and `~/.config/Code/User` will be a directory symlink to `config/Code/User`. All well and good, this is expected behavior. However, this is not how I want `~/.config/autostart` and `~/.config/Code/User` setup. I would like a means to be able to use the `config/*/*` globbing but also be able to specify a globbing pattern to exclude. This will allow me to specify more appropriate link actions for `config/autostart` and `config/Code/User` while keeping a more general catch all pattern with handles the majority of cases.

Here is a possible new version of `install.conf.yaml` using this exclusion:

```yaml
# defaults and clean are unchanged

- link:
    ~/.config/:
        path: config/*/*
        glob: true
        exclude: [config/autostart/*, config/Code/*]
#           - config/autostart/*
#           - config/Code/*
    ~/.config/autostart: config/autostart
    ~/.config/Code/User/:
         path: config/Code/User/*
         glob: true
```

From here, dotbot could use the patterns in `exclude` to generate a list of files using the same glob mechinisms that generates the targets list and then just perform set difference to get the final list. As a bonus, adding `exclude` paremeter does not break backwards compatibility :)

If there is no problem with this suggestion, I would like to implement this.

etkeys:
- Added `exclude` parameter to _link_. Now, an array of glob patterns
    can be given that will be used to remove items from a glob match.
    This parameter will only have an effect when `glob` is `true`.
- Updated README to add description for `exclude` and add in examples.

Resolves #247

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,7 @@ mapped to extended configuration dictionaries.
 | `glob` | Treat a `*` character as a wildcard, and perform link operations on all of those matches (default: false) |
 | `if` | Execute this in your `$SHELL` and only link if it is successful. |
 | `ignore-missing` | Do not fail if the source is missing and create the link anyway (default: false) |
+| `exclude` | Array of paths to remove from glob matches. Uses same syntax as `path`. Ignored if `glob` is `false`. (default: empty, keep all matches) |
 
 #### Example
 
@@ -218,6 +219,12 @@ Explicit sources:
       glob: true
       path: config/*
       relink: true
+      exclude: [ config/Code ]
+    ~/.config/Code/User/:
+      create: true
+      glob: true
+      path: config/Code/User/*
+      relink: true
 ```
 
 Implicit sources:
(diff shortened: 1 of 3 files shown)


anishathalye:
Apologies for the delay with this. Thank you for the PR!

PR #250, opened 2020-12-03
Make parsed cli options available to plugins

### LINKED ISSUE #249: As a plugin creator, I would like the ability to access what cli options were passed when dotbot was initial invoked.
### A: Plugin calls dotbot again as sub-process

Ref [dotbot-sudo](https://github.com/DrDynamic/dotbot-sudo). A plugin might have need to run dotbot again as a sub-process to do a cool thing. Some of the options passed when initially invoked could still be applicable in the sub-process as well (e.g. `--vebose` (#104), `--plugin-dir`, `--force-color`).

*dotbot-sudo* runs a script to deal with only those directives given under it, but I can imagine a scenario where I would want to run dotbot using a different config file. For example, *.dotfiles* repo also installs *.dotfiles_local*

### B: Plugin calls something else as a sub-process

To extend the previous section a bit more. What if a plugin calls some other application as a sub-process that can print output of varying levels. Let's say we want to pass a `--verbose` flag to this sub-process only if *dotbot* was called with `--verbose`.

## Case 2: Plugin handles mutually exclusive directives

I had a really good scenario for this, but then I found out that `--exclude` exists and then there's [cf36bb](https://github.com/anishathalye/dotbot/commit/cf366bbf6676d1c95f412eb514509f16322b5c9c)...so I'll just keep that to myself for now.

## Case 3: The future

Who know what the future holds, am I right? What if more cli options get added that plugins could leverage? I'll concede this point as it's unlikely that 1.x will implement such a cli option, but maybe something to consider for 2.0?

## Current workaround
From what I've been able to determine, a plugin could just examine `sys.argv` to determine what cli options were passed. By the time the plugin is used, the options have been validated and should be safe to use. This seems hacky though. The core application validates/manipulates the cli options to setup the appropriate context. A plugin shouldn't circumvent the work done by core and possibly damage the current context.

Also, what if using just `sys.argv` becomes improper? Either a new method is used or `sys.argv` is augmented with addition data? A plugin would be unaware of this condition.

etkeys:
- `Context` now provides a copy of the parsed cli options so that they
    can be used by plugins.
- Modified `Dispatcher` to pass parsed cli options to `Context`

Resolves #249

I'd consider myself an experienced junior developer with python. I'm questioning if I used optional arguments correctly. Also, I wasn't sure if optional arguments were to be listed alphabetically. I tried to do what I thought was best to not break backwards compatibility.

### DIFF
--- a/dotbot/cli.py
+++ b/dotbot/cli.py
@@ -97,7 +97,7 @@ def main():
             # default to directory of config file
             base_directory = os.path.dirname(os.path.abspath(options.config_file))
         os.chdir(base_directory)
-        dispatcher = Dispatcher(base_directory, only=options.only, skip=options.skip)
+        dispatcher = Dispatcher(base_directory, only=options.only, skip=options.skip, options=options)
         success = dispatcher.dispatch(tasks)
         if success:
             log.info('\n==> All tasks executed successfully')
--- a/dotbot/context.py
+++ b/dotbot/context.py
@@ -6,9 +6,10 @@ class Context(object):
     Contextual data and information for plugins.
     '''
 
-    def __init__(self, base_directory):
+    def __init__(self, base_directory, options):
         self._base_directory = base_directory
         self._defaults = {}
+        self._options = options
         pass
 
     def set_base_directory(self, base_directory):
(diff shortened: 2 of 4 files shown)


etkeys:
Currently experimenting with changing the default of value `options` when passed to `Dispatcher` constructor to a `Namespace` type object instead of `None`.

anishathalye:
Yeah I agree, seems good / harmless (since plugins could directly access argv anyways) to expose parsed options to plugins.

Made some small tweaks, squashed/rebased, and merged in 5f1e33ed67aebf.

PR #253, opened 2020-12-04
Add cli option force shell show stderr/stdout

### LINKED ISSUE #104: Verbose mode: provides command failure details
Even if I enable the verbose flag, no details are provided in case of command failure

etkeys:
Passing `--verbose` flag two times will now force shell commands to show stderr/stdout output regardless of settings in config file.

Resolves #104 

**Commits d6e1e4a and earlier are requested in PR #250 **
**Commits 33a4276 and later are dependent on changes above**

### DIFF
--- a/dotbot/cli.py
+++ b/dotbot/cli.py
@@ -1,7 +1,7 @@
 import os, glob
 import sys
 
-from argparse import ArgumentParser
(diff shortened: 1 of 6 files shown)


anishathalye:
Thank you! Merged 43b62ed53.

PR #255, opened 2020-12-26
Canonicalize destination path for relative-path: True

cdleonard:
The python os.path.relpath does not touch the filesystem and can return
an invalid result when the "start" argument contains symlinks.

This fixes using dotbot when $HOME is a symlink.

Signed-off-by: Leonard Crestez <cdleonard@gmail.com>

### DIFF
--- a/dotbot/plugins/link.py
+++ b/dotbot/plugins/link.py
@@ -188,6 +188,7 @@ class Link(dotbot.Plugin):
         Returns the relative path to get to the source file from the
         destination file.
         '''
+        destination = os.path.realpath(destination)
         destination_dir = os.path.dirname(destination)
         return os.path.relpath(source, destination_dir)
 


anishathalye:
Can you expand a bit on what you're doing, what the current behavior is, and what the expected/desired behavior is?

I also wonder how this should interact with `canonicalize: false`. In this case, should we skip canonicalization?

cdleonard:
It breaks for me on a machine where /home/leonard is a symlink to /mnt/data/home/leonard

```
$ python3 -m dotbot -c /mnt/data/home/leonard/files/work/private-scripts/install.conf.yaml
Removing ~/.config/mc/ini
Creating link ~/.config/mc/ini -> ../../../../mnt/data/home/leonard/files/work/private-scripts/mc.ini
```

The config is trivial:
```
- link:
    ~/.config/mc/ini:
      create: true
      if: 'which mc'
      path: mc.ini
```

With the patch the behavior is much saner:
```
$ python3 -m dotbot -c /mnt/data/home/leonard/files/work/private-scripts/install.conf.yaml
Removing ~/.config/mc/ini
Creating link ~/.config/mc/ini -> ../../files/work/private-scripts/mc.ini
```

The problem is that if the source is canonicalized without the destination then result can be incorrect. Should be both or neither.

cdleonard:
On second though putting "canonicalize: false" in my install.conf.yaml also fixes the issue, and this is probably better than fixing dotbot source.

cdleonard:
I'm not sure this is incorrect behavior from dotbot, I'll fix my issue some other way.

PR #268, opened 2021-04-16
Support both `[]` and `?` glob patterns

### LINKED ISSUE #267: link glob patterns should support more than just `*`
While the `glob.glob()` function supports more than just `*`, including `[]` character sets and `?` single character matches, the `glob: true` for links does not.  

Patch forthcoming.

eengstrom:
fixes #267


anishathalye:
Thank you for the fix and tests :)

Merged!

PR #272, opened 2021-04-16
add link `prefix: string` option

### LINKED ISSUE #259: Add `prefix:` option to links created with `glob: true`
Perhaps this is a niche case, but in my `dotfiles`, I have a **subdirectory** for most of the true "dot" files.  i.e.:

```
dotfiles
├── README.md
├── bin
│   ├── dotbot
│   ├── look
│   ├── pbfile
│   └── ...
├── dot
│   ├── bashrc
│   ├── gitconfig
│   ├── gitignore
│   ├── gorc
│   ├── login
│   ├── ...
│   ├── zshrc
│   └── zshenv
├── ...
```

I don't believe there is a way using the `link: ` module to create symlinks in my home directory for **all** the files in the `dot` subdirectory, sans excluded ones, given #248 when merged.

That is, what I want is to replace the MANY lines in my `dotconf.yml` like this:

```
- link:
    # Then everything in dot subdir:
    # Wish I could do this via a glob
    ~/.bashrc:          dot/bashrc
    ~/.gitconfig:       dot/gitconfig
    ~/.gitignore:       dot/gitignore
   ...
```
... with a glob version, something like this:

```
- link:
    ~/:
       glob: true
       path: ~/.dotfiles/dot/*
       prefix: '.'
```

Note the addition of the `prefix: ` option which specifies the prefix to be added to the basename of the file to be symlinked.  

Alternatives welcome, or if I missed something in the documentation for how this should be configured, a pointer to the documentation.

eengstrom:
/fixes #259 

Builds on previous commits / PRs (#268 and #271) which were needed to fix implicit some assumptions in the glob patterns and construction of the basename to which to prepend the prefix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,16 +181,22 @@ mapped to extended configuration dictionaries.
 | `force` | Force removes the old target, file or folder, and forces a new link (default: false) |
 | `relative` | Use a relative path to the source when creating the symlink (default: false, absolute links) |
 | `canonicalize` | Resolve any symbolic links encountered in the source to symlink to the canonical path (default: true, real paths) |
-| `glob` | Treat a `*` character as a wildcard, and perform link operations on all of those matches (default: false) |
 | `if` | Execute this in your `$SHELL` and only link if it is successful. |
 | `ignore-missing` | Do not fail if the source is missing and create the link anyway (default: false) |
-| `exclude` | Array of paths to remove from glob matches. Uses same syntax as `path`. Ignored if `glob` is `false`. (default: empty, keep all matches) |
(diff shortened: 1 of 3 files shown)


anishathalye:
This looks good, but I'll wait until #271 is resolved/merged before merging this. Do you think it would be helpful to expand a bit on the documentation in the README for this? Or working this into the full example? I think the main use case for this will be to `prepend: '.'` along with a glob, right?

anishathalye:
Thank you! One question, just to confirm, was the last `*` in `*files**` a stray asterisk (076939da8ee683d3cba791914c227f29c21fe042)?

eengstrom:
Yup - extra - thanks for the catch.

PR #275, opened 2021-04-28
Fixes typos in README "being" to "begin"

Justintime50:
Tiny PR to correct 2 typos in the README - "being" should be "begin".


anishathalye:
Thank you!

Issue #300, opened 2022-03-04
BUG: If the same destination is used multiple times with globs, only the last one is taken into account.


ldorigo:
Hi, I have two directories in my dotfiles: `/services` (systemd services that need to start on boot) and `/timers` (systemd timers with corresponding units). I want both to be linked into `~/.config/systemd/user`.

I tried doing:
```
- link:
    ~/.config/systemd/user: 
      path: services/*
      glob: true
    ~/.config/systemd/user:
      path: timers/*
      glob: true
``` 

But it looks like it only links the second one (timers), not the first. Any workaround? (I can't just put them all into the same folder because I am also starting them up with dotbot - for services I need to start/enable the .service file, and for timers only the `.timer` file).

ldorigo:
I also tried putting the three directories inside a common `systemd/` directory and  doing:

```
- link:
    ~/.config/systemd/user/: 
      glob: true
      path: systemd/**
```

But it doesn't work - it only links the directories, so `systemd/services` and `systemd/timers`, not the files in those directories. So the documentation is wrong (docs say "** matches any file, recursively").

And doing:

```
- link:
   ~/.config/systemd/user/: 
      glob: true
      path: systemd/**/*
```

Also doesn't work, it does link all the files but still within the folder structure (so it creates `~/.config/systemd/user/services/a.service`, `~/.config/systemd/user/services/b.service`,  etc.)

ldorigo:
I took a look at the source code, seems like this is due to a limitation of Yaml/Json - the destination paths are used as keys, and there can't be duplicate keys. This means it's not completely trivial to solve it, I see two possibilities:

1. Allow specifying lists of paths rather than a single path. This should be relatively easy to implement - Yaml and Json both support lists, all it would take is adding an additional case in the `Links` plugin. But I'm not familiar enough with the codebase to see possible pitfalls.
2. Add an additional option (something like "preserve_directory_tree") that lets you choose whether to link all files in subdirectories in a flat way vs. preserving the directory structure. 

I think (1) is better - (2) is very specific to this usecase and could be confusing, while (1) is quite explicit (you need to pass a list of directories by hand). If you approve I can submit a PR later.

ldorigo:
Wanted to get this working so I added a PR, let me know if there's any problems with it

anishathalye:
Related to this point: note to self, maybe we should add an explanation of this to the documentation, that the yaml `key: value` pairs don't support duplicate keys.

[...]

PR #301, opened 2022-03-04
Add the possibility of specifying more than one source path ...

ldorigo:
… in which case the destination is assumed to be a folder in which source paths are linked.

This is to solve #301. The diff on github is borked and shows lots of changes, but the changes are actually quite small:

- `Link._default_sources` was changed to return a list of sources rather than a single source
- `path` is changed to `paths` and all of the linking action is wrapped in an additional loop that iterates over source paths
- If globbing is active, additional tests are added for edge cases
- If there is no globbing and several paths are given, the destination path is assumed to be a directory, so we adapt the destination path accordingly

If you're ok with the approach I'll also edit the documentation. I don't really understand how the tests work, but I'll also try to add a few test cases when I have time to look more into it.

### DIFF
--- a/dotbot/plugins/link.py
+++ b/dotbot/plugins/link.py
@@ -52,87 +52,103 @@ class Link(dotbot.Plugin):
                 base_prefix = source.get("prefix", base_prefix)
                 ignore_missing = source.get("ignore-missing", ignore_missing)
                 exclude_paths = source.get("exclude", exclude_paths)
-                path = self._default_source(destination, source.get("path"))
+                paths = self._default_sources(destination, source.get("path"))
             else:
-                path = self._default_source(destination, source)
+                paths = self._default_sources(destination, source)
             if test is not None and not self._test_success(test):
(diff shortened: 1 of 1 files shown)


anishathalye:
Thanks for the PR! I haven't taken a close look at the diff/code yet, but my high level understanding is that this seems to be useful functionality that is backwards-compatible.

I'll take a closer look in a couple days.

If it's effort to get the tests running locally, don't worry about it, I can add a test based on your example in https://github.com/anishathalye/dotbot/issues/300#issuecomment-1059172312.

ldorigo:
Hey, I haven't had time to work more on this and I'm handing in my master's thesis soon so won't have time to look at it for a while more --- if you want to proceed with adding a couple  of tests and closing this issue, please don't wait on me :-)

anishathalye:
I would like to avoid this complexity, and instead suggest the approach specified in https://github.com/anishathalye/dotbot/issues/300#issuecomment-1493372564 for people who need this functionality.

PR #308, opened 2022-04-11
Fix Windows symlinks

### LINKED ISSUE #307: Windows symlinks are always recreated
On Windows, `os.readlink()` returns a UNC path that never matches the anticipated symlink destination. This causes dotbot to always recreate the links (example dotbot output below):

```
Incorrect link ~/.gitconfig -> \\?\C:\Users\Me\Documents\dev\dotfiles\gitconfig
```

I tested two possible fixes:

1. Use `os.path.realpath()` instead of `os.readlink()`
2. Conditionally remove the `\\?\` prefix when dotbot is running on Windows.

Option 1 results in a unit test failure:

```
[42/55] (link-relink-relative-leaves-file.bash)
relink relative does not incorrectly relink file
- test failed.
-> fail!
```

Therefore I recommend going with option 2, which gives expected results (example dotbot output below):

```
Link exists ~/.gitconfig -> C:\Users\Me\Documents\dev\dotfiles\gitconfig
```

kurtmckee:
On Windows, `os.readlink(path)` returns UNC paths that start with `\\?\`.

The UNC prefix can simply be removed when running on Windows.

Fixes #307

### DIFF
--- a/dotbot/plugins/link.py
+++ b/dotbot/plugins/link.py
@@ -196,7 +196,10 @@ class Link(dotbot.Plugin):
         Returns the destination of the symbolic link.
         """
         path = os.path.expanduser(path)
-        return os.readlink(path)
+        path = os.readlink(path)
+        if sys.platform[:5] == "win32" and path[:4] == "\\\\?\\":
+            return path[4:]
+        return path
 
     def _exists(self, path):
         """
--- a/dotbot/plugins/link.py
+++ b/dotbot/plugins/link.py
@@ -4,7 +4,6 @@ import glob
 import shutil
 import dotbot
 import dotbot.util
-import subprocess
 
 
 class Link(dotbot.Plugin):


anishathalye:
Thank you for the pull request! By the way, did you get the unit tests running on Windows? What did you need to do to make that happen?

kurtmckee:
My pleasure!

I was able to get the unit tests running with Windows as the _host_ -- it still ultimately runs in a VirtualBox VM that Vagrant set up. (And THANK YOU for excellent instructions for recreating the test environment! It allowed me to make changes and avoid breaking functionality in Linux!)

I looked at how I could make the tests actually run under Windows, but there were several challenges:

1. **Test isolation**
    It might be possible to redesign the tests to run in temp directories, rather than relying on true filesystem isolation. I didn't look closely at that yet, though.
2. **Testing across Python versions**
   I'm familiar with a tool called tox that does a fantastic job of setting up testing environments for different Python versions, matrices of dependencies, plugins, etc etc etc. Tox would be a good way to test everything locally instead of relying just on CI. I strongly recommend it, always.

If you're interested, I can take a look at updating the tests so they can run on Windows. I do want you to bless the attempt before I try, though. :grinning:

kurtmckee:
Windows used to require admin privileges to create symlinks, but that requirement is dropped if Developer Mode is enabled (or if group policy is modified, I think). Also, Python calls the create-symlink API (whatever it's called) with the correct arguments as of...2019, maybe? Hopefully those changes reduce the weirdness.

I'll try to get the tests working in Windows. I mentioned running the tests locally because I typically target locally-runnable tests, then make CI just run the same test suite. Local and CI testing are both valuable. :+1:

kurtmckee:
Closing because this work is superseded by the work associated with #309.

Issue #312, opened 2022-05-05
feature-request: run a command **IF** a condition passes


Clumsy-Coder:
just like **[LINK](https://github.com/anishathalye/dotbot#format)** directive, there should one for running individual commands for SHELL directive.

Ex:
```yaml
- shell:
    - command: apt update && apt upgrade -y
      if: lsb_release -i | grep -io 'debian'
      description: Update APT package repository

    - command: dnf update -y
      if: lsb_release -i | grep -io 'fedora'
      description: Update DNF package repository
```

## why
- provides a better flow of running dotbot process
- can contain similar commands in a single shell directive. 
  - Ex: shell directive for updating repositories
  - Ex: installing a package using package manager in one OS, downloading a script from another OS

anishathalye:
This is already supported (without special syntax) by having the `if` be part of the command itself. E.g. `if (lsb_release -i | grep -io 'debian'); then apt update && apt upgrade -y; fi`

Clumsy-Coder:
Going to work on making this feature.

Clumsy-Coder:
@anishathalye I created a pull request implementing this feature. Have a look

PR #321, opened 2022-08-12
feat(shell): run shell command conditionally

Clumsy-Coder:
## what
- Run shell command conditionally (optional)
- Update README docs

## how
- the shell directive will check if the 'if' property is available.
  - If it is, it will run the command in 'if' property.
  - If the command results in a return 0 code or 'true', then the 'command' property will run.
  - If the condition to run the command is a non 0 code or 'false', the shell command won't run
- dotbot config example:
```yaml
- shell:
    - command: echo "this is running on a MacOS"
      if: uname -s | grep -i "Darwin"
```
- running shell command regardless
```yaml
- shell:
    - command: echo "This should run regardless"
```

- dotbot config example: skipping command if false
```yaml
- shell:
    - command: echo "This command should be skipped"
      if: false
```

## why
- can run the shell command conditionally
- can use multiple dotbot configs to run different OS
  - Ex: incorporating it in https://github.com/ecarlson94/dotbot-template

## where
- file changed in `./dotbot/plugins/shell.py`

## usage
Create dotbot config
```yaml
- shell:
    - command: echo "this is running on a MacOS"
      if: uname -s | grep -i "Darwin"
```

```yaml
- shell:
    - command: echo "This command should be skipped"
      if: false
```

```yaml
- shell:
    - command: apt update && apt upgrade -y
      if: lsb_release -i | grep -io 'debian'
      description: Update APT package repository

    - command: dnf update -y
      if: lsb_release -i | grep -io 'fedora'
      description: Update DNF package repository
```

# NOTE
- The reason to wrap the if command in string, is to prevent Python from interpreting it.
  - `str(item["if"]),`
  - https://github.com/Clumsy-Coder/dotbot/blob/25ef5d5a5fed1e2f95e57115dfdbc8ba67cfc93f/dotbot/plugins/shell.py#L52
  - Try it without wrapping it in strings
    - https://github.com/Clumsy-Coder/dotbot/blob/974156ccdecee05eb9544b0b21cc0351428ddd43/dotbot/plugins/shell.py#L51-L57
    - ```python
      run_if_result = dotbot.util.shell_command(
          item["if"],
          cwd=self._context.base_directory(),
          enable_stdin=False,
          enable_stdout=False,
          enable_stderr=stderr,
      )
      ```
    - Fails to run because error was thrown
      ```yaml
      - shell:
          - command: echo "This command should be skipped"
            if: false
      ```
    - Works if wrapping `false` in strings
      ```yaml
      - shell:
          - command: echo "This command should be skipped"
            if: "false"
      ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,6 +332,7 @@ fine-grained control.
 | `stdin` | Allow a command to read from standard input (default: false) |
 | `stdout` | Show a command's output from stdout (default: false) |
 | `stderr` | Show a command's error output from stderr (default: false) |
+| `if` | Run command if a condition is true (default: true) (optional) |
 
 Note that `quiet` controls whether the command (a string) is printed in log
 output, it does not control whether the output from running the command is
@@ -354,6 +355,19 @@ printed (that is controlled by `stdout` / `stderr`). When a command's `stdin` /
   -
     command: read fail
     stderr: true
+```
+
+##### Running shell command conditionally
+
+```yaml
+- shell:
+    - command: apt update && apt upgrade -y
+      if: lsb_release -i | grep -io 'debian'
+      description: Update APT package repository
+
+    - command: dnf update -y
+      if: lsb_release -i | grep -io 'fedora'
+      description: Update DNF package repository
 ```
 
 ### Clean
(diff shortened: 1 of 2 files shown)


Clumsy-Coder:
Pull Request connected to issue #312

Clumsy-Coder:
I tested it as much as I can.

This is first time working in Python code.

So, if issue surfaces, let me know

wonderbeyond:
Hi, I just made the [dotbot-if](https://github.com/wonderbeyond/dotbot-if) plugin, and found it's sharing the similar idea with this PR.

Issue #325, opened 2022-10-01
Named conditions (and block directive)


gnfzdz:
There's actually two requests here, but given the overlap I thought it suitable to just raise a single issue:
1. Introduce a new type of plugin, condition plugins, that provide named, configurable conditions written in Python
2. Introduce a new directive accepting: condition(s) to evaluate and a list of tasks/actions to execute when true

## Motivation
There are currently 3 plugins linked from the official documentation that provide a condition and a container of tasks to be executed based on the outcome ([if](https://github.com/wonderbeyond/dotbot-if), [ifarch](https://github.com/ryansch/dotbot-ifarch), [ifplatform](https://github.com/ssbanerje/dotbot-ifplatform)).  Dotbot users have already adopted this type of plugin so the interest is clearly there.

The most flexible (and similar to this request) is the [if](https://github.com/wonderbeyond/dotbot-if) plugin which takes a single shell command. Personally, there are at least a couple other conditions that I'd like to share with the community that aren't well suited to a shell one-liner.

Notably, all three of the plugins also share the same issue. They don't propagate defaults from the parent context (and scope any new defaults set to the child context).

I think it would be great to provide improved support for this pattern in the core dotbot project itself.

## Examples
First, you can see below a couple options for a hypothetical new directive. I don't actually have a strong preference here. Whatever we think is least likely to conflict with a custom plugin name.
```yaml
- cond:
    if:
      tty: true
    then:
      - shell: '<command_that_may_prompt_for_input>'
    else:
      - shell: 'fallback_command'

- block:
    when:
      tty: true
    tasks:
      - shell: '<command_that_may_prompt_for_input>'
```

Next you can see a few variants of the condition. I actually think it makes sense to support all of these.
```yaml

# For a string, it is interpreted as a shell command (providing consistency with the link directive's current if option) 
- cond:
    if: "command -v ansible"
    then:
      - shell: 'echo "I found ansible in $PATH"'

# For a dict, use each key to find the corresponding Condition plugin and pass the value as arguments
- cond:
    if:
      tty: true
      platform: 'arch'
    then:
      - shell: 'echo "I am running in a TTY on ArchLinux!"'

# An array is also supported for the same reasons as with dotbot tasks/actions.
# It allows repeating a condition multiple times.
- cond:
    if:
      - decrypted: "./config/file/a"
      - decrypted: "./config/file/b"
      - 'command -v "program_requiring_sensitive_config"'
    then:
[...]

PR #313, opened 2022-05-18
Add support for Windows and Cygwin

kurtmckee:
This PR introduces the following changes:

* Add full support for Windows Powershell (CPython 3.8+) and Cygwin (CPython 2.7+, PyPy 2.7, and PyPy 3.9).
* Migrate all existing unit tests from bash to Python.
* Isolate all unit test filesystem interactions so they run entirely in a temporary directory.
  This includes verification that files are not being opened or created outside the temporary directory.
* Test dotbot in CI on Linux, MacOS, and Windows, across all supported CPython and PyPy versions.
* Add code coverage reports when running locally.
  This includes summary output at the CLI and generated HTML reports in `htmlcov/`.
* Document how to run the unit tests locally, including environment setup.

The changes have been tested on these platforms:

* Windows Powershell (locally, and in CI)
* Linux (locally, and in CI)
* Cygwin (locally)
* MacOS (in CI)

Closes #102
Closes #307
Closes #309
Closes #323

PR #326, opened 2022-10-01
Add generic conditional container plugin.

gnfzdz:
See #325 for explanation

Not intended as merge worthy, but just something to help visualize the idea. Let me know if it's worth refining.

### DIFF
--- a/dotbot/plugins/conditional.py
+++ b/dotbot/plugins/conditional.py
@@ -1,0 +1,43 @@
+import dotbot
+from dotbot.dispatcher import Dispatcher
+from dotbot.tester import Tester
+
+class Conditional(dotbot.Plugin):
+
+    '''
+    Conditionally execute nested commands based on the result of configured test(s)
+    '''
+
+    _directive = "conditional"
+
+    def can_handle(self, directive):
+        return directive == self._directive
+
+    def handle(self, directive, data):
+        if directive != self._directive:
+            raise ValueError("Conditional cannot handle directive %s" % directive)
+        return self._process_conditional(data)
+
+    def _process_conditional(self, data):
+        success = True
+        tests = data.get("if")
+        test_result = Tester(self._context).evaluate(tests)
+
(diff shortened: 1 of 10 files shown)


dotbot/plugins/conditional.py
```diff
@@ -0,0 +1,43 @@
...
+        tasks = data.get("then") if test_result else data.get("else")
+        self._log.info("executing sub-commands")
+        # TODO prepend/extract defaults if scope_defaults is False
+        if tasks is not None:
+            return self._execute_tasks(tasks)
+        else:
+            return True
+
+    def _execute_tasks(self, data):
+        # TODO improve handling of defaults either by reusing context/dispatcher -OR- prepend defaults & extract at end
```
gnfzdz, line 35: Note this currently has the same issue as the similar third-party plugins (scoping defaults to the child tasks). I think the best approach would be to update dispatcher constructor to allow passing in (and therefore reusing) a context.

Other options considered:
- Reusing the parent dispatcher by exposing it through context
- Modify the task list by prepending a default directive. After completion, get the child context defaults and call set_defaults on parent context.

In either case, I also think it's worth adding a configuration option to this plugin to control whether defaults are scoped to the child tasks or not.

kurtmckee:
This is a great start, @gnfzdz! I'd like to assist on implementing a number of initial condition providers, writing up documentation, and adding unit tests. I can pick this up after #313 is reviewed and merged, as it will dramatically affect how the unit tests are implemented (and will make it possible to test the condition providers across all supported operating systems).

gnfzdz:
@kurtmckee Thanks for the positive feedback. I can definitely see how this would work better with cross platform support. Thanks for the offer to help?

@anishathalye Have you had a chance to review the proposal? I did release this as a [separate plugin](https://gitlab.com/gnfzdz/dotbot-conditional), but it would feel a lot cleaner if at least the abstract base Condition was available in the core Dotbot project.

PR #327, opened 2022-10-26
don't show command if `quiet: true` and description is not set

### LINKED ISSUE #265: Shell commands without description cause the command to be shown even if `quiet: true`
Hi, thank you for maintaining such a great tool.

Currently, Dotbot shows the command if its `quiet` option is set to `true` but no description. Is it intended behavior?

For example, with this config, Dotbot will only show the command's stdout:

```install.conf.yaml
- shell:
    - command: echo 'Hello, world!'
      description: ''
      quiet: true
      stdout: true
```

```
% ./install

Hello, world!
All commands have been executed

```

but with this one, it will also show the command itself:

```shell
- shell:
    - command: echo 'Hello, world!'
      quiet: true
      stdout: true
```

```
% ./install
echo 'Hello, world!'
Hello, world!
All commands have been executed

```

s1as3r:
## Why
I have a shell directive and only want to see the command's output. I set `quiet: true` and `stdout: true`. When I ran dotbot, It showed the command, which I think should not be the case.

Resolves #265

### DIFF
--- a/dotbot/plugins/shell.py
+++ b/dotbot/plugins/shell.py
@@ -42,10 +42,11 @@ class Shell(dotbot.Plugin):
             else:
                 cmd = item
                 msg = None
-            if msg is None:
+            if quiet:
+                if msg is not None:
+                    self._log.lowinfo("%s" % msg)
+            elif msg is None:
                 self._log.lowinfo(cmd)
-            elif quiet:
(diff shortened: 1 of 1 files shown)


anishathalye:
Fixed the tests; merged in 9593510. Thank you!

PR #332, opened 2023-02-19
Fix dispatcher when no plugins are given

Wanimatrix:
When any plugin is using the dispatcher for nested actions it should
pass all plugins again to avoid that any plugin as subtask would be
handled.
Currently this means a lot of plugins are broken because they don't pass
plugins to the dispatcher.

To fix this, take all Plugin subclasses when no plugin is given (we
should at least have the default plugins).

### DIFF
--- a/dotbot/dispatcher.py
+++ b/dotbot/dispatcher.py
@@ -18,7 +18,10 @@ class Dispatcher(object):
     ):
         self._log = Messenger()
         self._setup_context(base_directory, options)
-        plugins = plugins or []
+        if plugins == None:
+            plugins = Plugin.__subclasses__()
+        else:
+            plugins = plugins or []
         self._plugins = [plugin(self._context) for plugin in plugins]
         self._only = only
         self._skip = skip
--- a/tests/dotbot_plugin_dispatch.py
+++ b/tests/dotbot_plugin_dispatch.py
@@ -1,0 +1,17 @@
+"""Test that a plugin can call dispatcher for subtasks.
+
+The plugin calls dispatch with his data.
+"""
+
+import os.path
+
+import dotbot
+
+
+class Dispatch(dotbot.Plugin):
+    def can_handle(self, directive):
+        return directive == "dispatch"
+
+    def handle(self, directive, data):
+        dispatcher = dotbot.dispatcher.Dispatcher(self._context.base_directory())
+        return dispatcher.dispatch(data)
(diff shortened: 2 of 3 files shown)


tests/dotbot_plugin_dispatch.py
```diff
@@ -0,0 +1,17 @@
+"""Test that a plugin can call dispatcher for subtasks.
+
+The plugin calls dispatch with his data.
```
kurtmckee, line 3:
```suggestion
The plugin calls dispatch with its data.
```

dotbot/dispatcher.py
```diff
@@ -18,7 +18,10 @@ def __init__(
     ):
         self._log = Messenger()
         self._setup_context(base_directory, options)
-        plugins = plugins or []
+        if plugins == None:
```
kurtmckee, line 21: Please always use identity, not equality, comparisons with `None`:

```suggestion
        if plugins is None:
```

dotbot/dispatcher.py
```diff
@@ -18,7 +18,10 @@ def __init__(
...
+            plugins = Plugin.__subclasses__()
```
kurtmckee, line 22: This will circumvent the configuration option `--disable-built-in-plugins`. It will re-enable plugins that the user may have requested be disabled.

kurtmckee (commented):
I don't think that this design will respect the `--disable-built-in-plugins` option.

kurtmckee:
I've submitted #343 to address this, but I think that it will respect the `--disable-built-in-plugins` option by simply returning the same Dispatcher instance after it is first instantiated.

As a side effect, #343 introduces a cached Dispatcher instance that plugins can begin leveraging -- they can simply access `dotbot.dispatcher.current_dispatcher` directly instead of re-instantiating the Dispatcher class.

kurtmckee:
@anishathalye This might be resolved by your recent changes.

anishathalye:
Not exactly, it still leaves old plug-ins broken, though it is more clear how to fix those plugins now.

We could fix the behavior with old plugins by changing how the Dispatcher with `plugins=None` behaves---rather than treating that as no plugins, we could have it use the same list of plugins that `cli.py` computes. I think that's a better fix / actually fixes backwards compatibility. Implemented here: https://github.com/anishathalye/dotbot/commit/def05688c02a2cbb540e5fcaaf31b369ccc18043

Thank you all for your input here.

Issue #284, opened 2021-07-23
Add Option to Ignore "Globbing couldn't find anything matching"


patbakdev:
I have the following config with the intention of being able to just drop in files/folders and have them automatically picked up. But if some folders are empty, then I get the warning ("Globbing couldn't find anything matching ...") and exit code 1. I'd like to have an option to ignore this warning and return exit code 0.

```
- defaults:
    link:
      create: true
      relink: true
      glob: true

- clean:
    - ~/
    - ~/.config
    - ~/.local
    - ~/.local/bin

- link:
    ~/:
      path: ./.*
      exclude:
        - ./.config
        - ./.local
    ~/.config/:
      path: ./.config/*
    ~/.local/:
      path: ./.local/*
      exclude:
        - ./.local/bin
    ~/.local/bin:
      path: ./.local/bin/*
```

Perhaps something like:
```
- link:
    ~/:
      path: ./.*
      ignore_empty_glob: true     <-- NEW
      exclude:
        - ./.config
        - ./.local
```

brutus:
Same here. I'll add a quick PR (not thought through, missing test, etc. but works for me - so far YMMV).

anishathalye:
This is now unconditionally allowed (no `ignore_empty_glob: true` required), as of 4daa065.

PR #336, opened 2023-04-02
respect `ignore-missing` when using globs

brutus:
Not sure if clean or helpful, but works for me so far… see https://github.com/anishathalye/dotbot/issues/284


anishathalye:
Thanks for the PR. I prefer not to associate a dual meaning with `ignore-missing` (which creates links _anyways_ even if the source doesn't exist). Instead, 4daa065 unconditionally allows globs with no matches.

Issue #338, opened 2023-07-13
Drop support for python 2.7


aqeelat:
I would like to work on this issue

anishathalye:
What do you mean by dropping support for Python 2.7? The code is already Python 2.7 compatible, without much weird compatibility code. I don't want to remove functionality without good reason.

In fact, adding back Python 2.7 testing to CI would be ideal (I got lazy in 53b3781fbb6d926ada37f511b21897290df93df8).

kurtmckee:
I agree with this request, insofar as the project documentation and test suite drop all claims for Python 2.7 support.

The loss of Python 2.7 support in CI needs to be a clear death knell, not a challenge to find an alternate solution. Unless CI support can be restored, the project needs to drop claims of support for Python 2.7 for lack of compatibility testing on incoming PRs.

anishathalye:
Added back Python 2.7 testing in 712b30a44508dacba8d2f250753035f168e0812b. Is there anything else related to Python 2.7 support that we need to discuss, or should we close this issue?

kurtmckee:
It is possible to document that "dotbot 1.19.2 / commit 712b30a is the last version to officially support Python 2.7" and then allow the project to move forward.

Ongoing Python 2.7 support feels like it's a point of pride. I recommend deciding and documenting whether Python 2.7 will always be supported, or under what conditions dotbot will stop targeting Python 2.7 (like "_X_ years after 2.7 EOL", or "after _Y_ distro is EOL", etc). That will help users and developers understand the situation, and will make it easier to address tickets like this.

This ticket can be closed, but please consider making a decision about Python 2.7 and documenting it. I hope that the above will be taken as a recommendation, not a condemnation. I really do appreciate your work on dotbot!

anishathalye:
> Ongoing Python 2.7 support feels like it's a point of pride.

It's also a practical matter for me — I use many machines, and some of them are running old OSes with only Python 2 installed.

I don't want to _remove_ Python 2.7 support without good reason (the support is already there, and there is no burden for leaving it in at the moment). I think a good time to consider removing Python 2.7 support would be whenever a substantial new feature is being updated where Python 2.7 compatibility is difficult. Otherwise, I think there is no hurry.

kurtmckee:
> I use [...] old OSes with only Python 2 installed.

That helps clarify the ongoing support. Thanks!

PR #340, opened 2023-07-19
deps: update to use pyyaml 6

chenrui333:
relates to https://github.com/Homebrew/homebrew-core/pull/136978
also relates to https://github.com/yaml/pyyaml/issues/723 https://github.com/yaml/pyyaml/issues/601

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,7 @@ setup(
         "wheel>=0.31.0",
     ],
     install_requires=[
-        "PyYAML>=5.3,<6",
+        "PyYAML",
     ],
     extras_require={
         "dev": {


kurtmckee:
PyYAML 6 dropped support for Python 2.7, which dotbot still supports.

anishathalye:
This might be the [reason](https://github.com/anishathalye/dotbot/issues/338#issuecomment-1637786129) to drop support for Python 2.7, when our dependencies no longer support it (and upgrading dependencies is the easiest solution to problems like that posted in the OP).

c-c-k:
Hi, sorry for barging in and possibly being a noob.

Wouldn't dropping support for Python 2.7 be a good point to mark dotbot as dotbot 2, leaving a separate dotbot 1 branch in it's current state in case support for Python 2.7 becomes a high demand issue latter on?

Also in that case, maybe make Python 3.8 the minimal version for dotbot in order to start using pathlib?

Sorry again for being a bother and saying stupid things.

kurtmckee:
> Python 3.8 [...] in order to start using pathlib

[The `pathlib` module was introduced in Python 3.4](https://docs.python.org/3/library/pathlib.html). Python 3.8 is the minimum version needed for Windows support, however, due to its support for non-admin symlink creation on NTFS filesystems.

> to resolve the actual pyyaml issue in dotbot 1, maybe drop support for YAML

I think this introduce a new issue -- all existing dotbot users' configuration files would be rendered unusable.

c-c-k:
> [The `pathlib` module was introduced in Python 3.4](https://docs.python.org/3/library/pathlib.html). Python 3.8 is the minimum version needed for Windows support, however, due to its support for non-admin symlink creation on NTFS filesystems.

Sorry and thank you, I first noticed the pathlib module when I was trying to do something for a Win XP machine so I probably got it wrong then.
After looking over pathlib's documentation again I see that my suggestion in this regard is no good since pathlib keeps getting new features even after Python 3.8.

 > > to resolve the actual pyyaml issue in dotbot 1, maybe drop support for YAML
> 
> I think this introduce a new issue -- all existing dotbot users' configuration files would be rendered unusable.

What I mean is that dotbot 2 will become the default and dotbot 1 will need to be specifically selected by users that need pyhon2.7 support.
And going on the somewhat optimistic notion that users needing Python 2.7 support would  usually own another machine that can run a newer Python: Perhaps add a script that would automatically convert yaml config files to json ones, so a user could run the auto-conversion script on the Python 3.x machine and then push/pull the repo with the automatically generated json config files to the Python 2.7 machine.

Also maybe to reduce the naming confusion, maybe branch off "dotbot 2.7" as the legacy python 2.7 supporting version and keep the main dotbot version as simply "dotbot".
As a transition strategy maybe add a warning to the main (non python 2.7) dotbot version that it is no longer being tested for python 2.7 support and is not guaranteed to  work on it.

kurtmckee:
I don't think that this discussion is furthering this PR. Let's end the discussion here for now.

c-c-k:
Sorry, I have deleted the offending comments.

johnlettman:
I advocate for discontinuing support for Python 2.7 due to several reasons:

1. **End-of-Life Status**: Python 2.7 has been [beyond its end-of-life][pyclock] for [about three years now][pyversions], following its last [extension][py2.7extension] after the official [sunsetting][py2sunset]. Consequently, Python 2 is no longer available for [download][pydownloads] on the Python website.
2. **Trend in Python Projects:** Numerous prominent frameworks, projects, and packages, such as [OpenStack][o7kpy3], have already [transitioned to Python 3][py3statement], leaving Python 2 behind.
3. **Improved Features:** Python 3 offers [several enhancements][py3features] that can significantly enhance development. Notably, the [typing features][py3typing] have proven valuable in my personal experience.

and, of course

4. **Security:** Dotbot, its plugins, and the users can benefit from the usage
   of supported Python interpreters and latest dependency updates through the
   switch. It can resolve the following CVE:
   
   **Improper Input Validation in PyYAML** / [`CVE-2020-14343`](https://nvd.nist.gov/vuln/detail/CVE-2020-14343) / [`GHSA-8q59-q68h-6hv4`](https://github.com/advisories/GHSA-8q59-q68h-6hv4) / *A vulnerability was discovered in the PyYAML library in versions before 5.4, where it is susceptible to arbitrary code execution when it processes untrusted YAML files through the full_load method or with the FullLoader loader. Applications that use the library to process untrusted input may be vulnerable to this flaw. This flaw allows an attacker to execute arbitrary code on the system by abusing the python/object/new constructor. This flaw is due to an incomplete fix for [CVE-2020-1747](https://github.com/advisories/GHSA-6757-jp84-gxfx).*

[pyclock]: https://pythonclock.org/
[pyversions]: https://devguide.python.org/versions/
[py2sunset]: https://www.python.org/doc/sunset-python-2/
[py2.7extension]: https://hg.python.org/peps/rev/76d43e52d978
[pydownloads]: https://www.python.org/downloads/
[py3statement]: https://python3statement.org/
[o7kpy3]: https://wiki.openstack.org/wiki/Python3

[py3features]: https://www.asmeurer.com/python3-presentation/python3-presentation.pdf
[py3typing]: https://docs.python.org/3/library/typing.html

johnlettman:
I reckon the README already suggests no Python 2:
https://github.com/anishathalye/dotbot/blob/712b30a44508dacba8d2f250753035f168e0812b/README.md?plain=1#L70-L71

Anyways, I would be happy to help out with it too.

rht:
I am not able to install docbot properly via pip on Python 3.11. The reasons is the <6 pyyaml version constraint. pyyaml supported Python 3.11 only recently: https://github.com/yaml/pyyaml/commit/957ae4d495cf8fcb5475c6c2f1bce801096b68a5. You have to choose to support either of 3.11 or 2.7. I'd vote for supporting 3.11 and dropping EOL 2.7.

chenrui333:
> PyYAML 6 dropped support for Python 2.7, which dotbot still supports.

python 2.7 has been EOL for quite some time, and macOS has not shipped it by default either, also all package managers stop shipping packages using py2, so it would be good to just drop it and then prep for the upcoming py3.12 release :) just my two cents.

anishathalye:
Thank you for the PR. I didn't just want to remove the version specification altogether; alternative change in https://github.com/anishathalye/dotbot/commit/b732baf163f03a82961f68db7a9327e8d4d9fb18 (you're credited in there).

Thanks all for your patience. This is included in Dotbot 1.20.0 and released on PyPI.

PR #343, opened 2023-07-28
Make Dispatcher a singleton after first instantiation

### LINKED ISSUE #339: All plugins using sub-directives implemented with Dispatcher broken
First, thanks for this great dotfiles manager! I recently migrated from my own >10y old semi-homebrewn solution with many clumsy shell scripts. Dotbot made it all much simpler for me! However there's currently a bug preventing some plugins from working...

As first reported at https://github.com/wonderbeyond/dotbot-if/issues/1#issuecomment-1532139808, it seems like all plugins that allow using arbitrary sub-expressions broke as of refactorings introduced in https://github.com/anishathalye/dotbot/commit/b5499c7dc5b300462f3ce1c2a3d9b7a76233b39b.

Let's take for example the [dotbot-if](https://github.com/wonderbeyond/dotbot-if) plugin that allows doing things like

```yaml
- if:
    cond: 'true'
    met:
    - shell:
      - echo The condition was true
```
The `met:` allows any dotbot directive to be run, here`shell` was used as the examples.

However as of the changes in https://github.com/anishathalye/dotbot/commit/b5499c7dc5b300462f3ce1c2a3d9b7a76233b39b where the Dispatcher was modified to no longer load all plugins by itself (used to be done with the helper `_load_plugins`), plugins relying on using the Dispatcher to parse and handle sub-directives does no longer work as other plugins are not loaded anymore. The plugins array is now an argument that will be passed in. The problem is that the plugins using the Dispatcher does not know all the other plugins to be loaded as this was done before in an outer context not available here in the plugin. The example above will result in 

```console
$ ./install -v
Action shell not handled

==> Some tasks were not executed successfully
```
as reported in https://github.com/wonderbeyond/dotbot-if/issues/1. **Please see this issue for details on how the mentioned refactor broke plugins using the Dispatcher.**

Now we could say that the problem is not dotbot but the plugins, and maybe it is so that all plugins should update. But to what, what method should plugins use to parse sub-directive now? However as this did work before and no new major breaking version of dotbot was released, I would consider the problem being in dotbot and not the plugins right now.

I could be very wrong in my analysis, and maybe there's already a way in which plugins like dotbot-if could execute sub-directives with the Dispatcher or another class. Let's figure this out together, as plugins like `if` is kind of essential for dotbot!

### LINKED ISSUE #357: Duplicate directives when custom plugins import builtins
If a plugin is added that imports `dotbot.plugins` (such as [dotbot-if](https://github.com/wonderbeyond/dotbot-if/blob/28b2fc6042eff75b4190cccf70357e4b8346cf76/if.py#L8)) the built-in plugins will be included multiple times and then duplicate execution of the directives.

Example (tested with dotbot version 1.20.1):
With [dotbot-if](https://github.com/wonderbeyond/dotbot-if/tree/main) in `dotbot-plugins` directory, and the following config:
```yaml
- shell:
    - description: Demo shell
      command: echo "Should run once"
```

```
$ dotbot/bin/dotbot -c test.conf.yaml -p dotbot-plugins/dotbot-if/if.py
Demo shell [echo "Should run once"]
All commands have been executed
Demo shell [echo "Should run once"]
All commands have been executed

```

Perhaps the Dispatcher should dedupe the plugin list, or plugin module loading could ignore builtins?

kurtmckee:
This allows plugins to re-instantiate the Dispatcher (or, now, use the instance cached in `current_dispatcher`) without knowing what plugins have been loaded by dotbot.

This fixes plugins that want to dispatch to other plugins.

Fixes #339
Fixes #357

### DIFF
--- a/dotbot/dispatcher.py
+++ b/dotbot/dispatcher.py
@@ -5,8 +5,23 @@ from .context import Context
 from .messenger import Messenger
 from .plugin import Plugin
 
+current_dispatcher = None
+
 
 class Dispatcher:
+    def __new__(cls, *args, **kwargs):
+        # After dotbot instantiates this class, the instance will be cached.
+        # Subsequent instantiations (say, by plugins) will return the same instance.
+        # This is needed because plugins don't have access to the entire configuration
+        # (for example, they won't know which plugins have been loaded).
+        # This ensures a consistent configuration is used.
+        global current_dispatcher
+        if current_dispatcher is None:
+            instance = object.__new__(cls)
+            instance.is_initialized = False
+            current_dispatcher = instance
+        return current_dispatcher
+
     def __init__(
         self,
         base_directory,
@@ -16,6 +31,9 @@ class Dispatcher:
         options=Namespace(),
         plugins=None,
     ):
+        if self.is_initialized:
+            return
+        self.is_initialized = True
         self._log = Messenger()
         self._setup_context(base_directory, options)
         plugins = plugins or []
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -12,6 +12,7 @@ import pytest
 import yaml
 
 import dotbot.cli
+import dotbot.dispatcher
 
 
 def get_long_path(path):
@@ -312,3 +313,9 @@ def run_dotbot(dotfiles):
             dotbot.cli.main()
 
     yield runner
+
+
+@pytest.fixture(autouse=True)
+def reset_current_dispatcher():
+    yield
+    dotbot.dispatcher.current_dispatcher = None
--- a/tests/test_dispatcher.py
+++ b/tests/test_dispatcher.py
@@ -1,0 +1,22 @@
+import dotbot.dispatcher
+
+
+def test_dispatcher_instantiation(home, dotfiles, run_dotbot):
+    """Verify that the dispatcher caches itself as a singleton."""
+
+    assert dotbot.dispatcher.current_dispatcher is None
+
+    dotfiles.write_config([])
+    run_dotbot()
+
+    # Verify the dispatcher has been cached.
+    assert dotbot.dispatcher.current_dispatcher is not None
+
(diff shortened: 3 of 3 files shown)


kurtmckee:
Test suite is failing due to PyYAML 5 and Cython 3. There is currently no way to trivially resolve this.

erikw:
@kurtmckee many thanks! I can confirm that with this PR, plugins like dotbot-if works (https://github.com/wonderbeyond/dotbot-if/issues/1) 🚢 :shipit:

kurtmckee:
_Rebased on `master` and resolved merge conflicts. The test suite passes now._

erikw:
I would be delighted if we could get this merged!

kurtmckee:
@anishathalye Friendly ping. :grin:

kurtmckee:
Fixes #357

kurtmckee:
Hello @anishathalye! I hope you're well.

Would you take a look at this PR? It includes a test case and has been confirmed to resolve issues in some dotbot plugins.

anishathalye:
Thanks for the reminder. I don't think this is quite the right solution—if there are plugins using the Dispatcher in any way other than "get the same behavior as the original Dispatcher", e.g., supplying different options to the constructor like a different base directory or supplying a custom plugin list, this will break them. Also, it's been quite a while since we broke the old behavior, so I'm worried about reverting it back to the old behavior and breaking something else. Instead, I exposed the set of plugins through the context, so that plugins can create a Dispatcher like the original. For more, see the commit message here: https://github.com/anishathalye/dotbot/commit/8c2dc8cbc651957c21176bb5eb5e709c31f3af2e.

Also, this PR doesn't actually fix #357 (see the regression test I added in https://github.com/anishathalye/dotbot/commit/cef40c902ef0f52f484153413142b5154bbc5e99, still fails with this PR).

PR #345, opened 2023-09-10
Fix wrong pyyaml path

bennyyip:
pyyaml 6.0.1 no longer use path `lib3`.


anishathalye:
Good catch, thank you!

PR #346, opened 2023-09-10
Support Python 3.12

kurtmckee:
All tests pass under Python 3.12 already, but the test suite was using a `shutil.rmtree` option, `onerror`, that Python 3.12 introduced a replacement for (`onexc`) and simultaneously deprecated.

This is addressed, pytest is now configured to escalate warnings into errors, and the test suite -- and CI -- have been updated to include Python 3.12 in their runs.


anishathalye:
Thank you!

PR #347, opened 2023-09-10
Use pre-commit

kurtmckee:
git pre-commit hooks can catch linting issues on developer PCs before they're ever committed to the repo. [pre-commit](https://pre-commit.com/) is a widely-used tool that allows projects to specify all of the linters they want to run before each commit.

This PR introduces a pre-commit config that will enforce standard text linting; black, isort, and editorconfig linting; and idiomatic Python 3.6+ syntax.

* Developers with the pre-commit tool installed can simply run `pre-commit install` in the dotbot repo, and from that point forward git will run the tool prior to each commit so the files are linted and auto-fixed.
* The dotbot repo can be registered with the [pre-commit.ci](https://pre-commit.ci/) service so that it auto-checks -- and, by default, autofixes! -- incoming PRs. It will also open PRs to keep the pre-commit hook versions updated on a regular basis. _I STRONGLY recommend doing this!_

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,33 @@
+# Run `pre-commit autoupdate` to update tool versions.
+# Note that pre-commit.ci can update hook versions automatically.
+
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks.git
+    rev: v4.4.0
+    hooks:
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: mixed-line-ending
+      - id: trailing-whitespace
+
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v3.10.1
+    hooks:
+      - id: pyupgrade
+        name: "Enforce Python 3.6 idioms"
+        args: [--py36-plus]
+
+  - repo: https://github.com/psf/black
+    rev: 23.9.0
+    hooks:
+      - id: black
+
+  - repo: https://github.com/pycqa/isort
+    rev: 5.12.0
+    hooks:
+      - id: isort
+
+  - repo: https://github.com/editorconfig-checker/editorconfig-checker.python
+    rev: 2.7.2
+    hooks:
+      - id: editorconfig-checker
--- a/dotbot/cli.py
+++ b/dotbot/cli.py
@@ -99,7 +99,7 @@ def main():
                 hash_msg = " (git %s)" % git_hash[:10]
             except (OSError, subprocess.CalledProcessError):
                 hash_msg = ""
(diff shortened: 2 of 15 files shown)


kurtmckee:
If unfamiliar with pre-commit, you can do the following to test it out:

* Checkout this branch
* Install pre-commit (for example, in an activated virtual environment using `pip install pre-commit` or using pipx -- which I strongly recommend -- using `pipx install pre-commit`)
* Enable pre-commit in the dotbot repo (run `pre-commit install`)
* Modify two files by rearranging their imports.
* Stage one of the two files (but not both) for a commit.
* Try to commit the file.

What you'll see is that pre-commit runs isort against the one staged file (but not both!), isort lints and fixes the file's imports, and the commit fails. Notably, pre-commit stashes changes temporarily so only what you've staged gets checked.

For example, this is the output I see locally when I rearrange the imports in `cli.py` and another file, but only stage `cli.py` in a commit:

```text
[WARNING] Unstaged files detected.
[INFO] Stashing unstaged files to C:\Users\Kurt\.cache\pre-commit\patch1694373671-15164.
check yaml...........................................(no files to check)Skipped
fix end of files.........................................................Passed
mixed line ending........................................................Passed
trim trailing whitespace.................................................Passed
Enforce Python 3.6 idioms................................................Passed
black....................................................................Passed
isort....................................................................Failed
- hook id: isort
- files were modified by this hook

Fixing C:\Users\Kurt\Documents\dev\dotbot\dotbot\cli.py

Check .editorconfig rules................................................Passed
[INFO] Restored changes from C:\Users\Kurt\.cache\pre-commit\patch1694373671-15164.
```

I have to stage isort's fixes to my commit before I can successfully commit.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: mixed-line-ending
+      - id: trailing-whitespace
+
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v3.10.1
+    hooks:
+      - id: pyupgrade
+        name: "Enforce Python 3.6 idioms"
```
henworth, line 17: Wouldn't this be counter to the desire to maintain compatibility with 2.7?

kurtmckee: Support for Pythons less than 3.6 was dropped ~2 weeks ago.

kurtmckee:
Hi Anish! I hope your year is off to a good start!

I'm going to close this PR to reduce cognitive load.

Thanks for your work on dotbot!

PR #356, opened 2024-02-21
Have a release pipeline

pirafrank:
Hi, this PR to add a pipeline to create releases with code and bundled artifacts.

There are a couple of reasons to do this:

- have a single (albeit fat) executable you can ship
- offer yet another way to run `dotbot`
- avoid dependence on Python on the target environment (many reasons to do this)

The pipeline currently builds and creates a release only on tag push.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,80 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - "v*.*.*"
+
+jobs:
+  release:
+    env:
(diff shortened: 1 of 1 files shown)


anishathalye:
Appreciate the pull request! The project packaging was substantially changed in abecc97bad09f0ef662787c85381ae8d3619224e, so I went with a different implementation to set up the release pipeline (still using pyinstaller, but building on the existing pyproject.toml etc.) in a514b8fb7fd13fbfc43770f3f3127f48a3b779be, crediting you in the commit message for the idea.

PR #372, opened 2025-01-14
Support creating hardlinks

### LINKED ISSUE #334: Hardlink support
There are some oddballs out there like Windows Terminal, which refuses to detect changes in the config file if it's a symlink (https://github.com/microsoft/terminal/issues/1812#issuecomment-508280785).

Recreating the symlink doesn't seem to do it either. The only currently viable solution is to use hardlinks, so it would be nice if dotbot can support it.

kurtmckee:
Closes #334

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,8 @@ can also be configured via setting [defaults](#defaults).
 
 ### Link
 
-Link commands specify how files and directories should be symbolically linked.
+Link commands specify how files and directories should be linked.
+Symlinks are created by default, but hardlinks are also supported.
 If desired, items can be specified to be forcibly linked, overwriting existing
 files if necessary. Environment variables in paths are automatically expanded.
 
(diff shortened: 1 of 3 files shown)


dgiagio:
Hi @anishathalye, have you had a chance to look at this PR? Is there anything else you need to merge it?
Thanks.

anishathalye (commented):
Two cases that this code doesn't handle:

- If a file already exists but is a symlink, and the user intends to create a hard link, this code skips over it with a "link exists" message, whereas it should show something like a "<file> already exists but is a symbolic link, not a hard link".
- If a file already exists and is a hard link pointing to the correct place, this code shows an error "<file> already exists but is a regular file or directory", whereas it should skip over it with the usual "Link exists ..." message—idempotence is currently broken.

Could you fix these two issues / add tests for them?

kurtmckee:
Not a fan of Copilot.

I'll see what I can do to address those concerns, but it's going to take a bit because life is busy currently!

kurtmckee:
@anishathalye I've addressed the cases you suggested.

In addition, I attempted to make the complex warning-message-handling logic clearer by rearranging and regrouping the conditions.

It may be easier to read outside of this PR's diff context:

https://github.com/kurtmckee/pr-dotbot/blob/hardlinks-issue-334/src/dotbot/plugins/link.py#L278-L323

anishathalye:
Thanks! There was a small bug with the refactor to `_link()`, which tried creating the link first and showed an error otherwise—it was using `os.exists()`, but we need `os.lexists()` to detect broken symbolic links and show the correct error message (the consequence was that we got a "Link failed" error instead of an "Invalid link" error). In the interest of time, I added a test and fixed this small bug.

Squashed in 5614642 and merged in 64d2330b2ee139f5c9603d10f19cadb07c65c234, thank you for submitting this pull request!

PR #373, opened 2025-01-14
Avoid deleting the source file when linking

### LINKED ISSUE #360: Accidental file deletion when linking
While configuring my dotbot config, I ran into an issue that caught me off guard and led to the accidental deletion of _source_ files.

```
# Directory structure
dotbot.yml
t_source/
------> /config.json
```

````yaml
# Initial dotbot.yml
- defaults:
    link:
      relink: true
      create: true
      force: true

- link:
    t_target: t_source
````

After running `dotbot -c dotbot.yml`, I get the following

```
$ dotbot -c dotbot.yml
Creating link t_target -> /home/user/test/t_source
All links have been set up

$ ls -l
-rw-rw-r-- 1 user user  110 Mar 21 22:48 dotbot.yml
drwxrwxr-x 2 user user 4096 Mar 21 22:47 t_source/
lrwxrwxrwx 1 user user   27 Mar 21 22:50 t_target -> /home/user/test/t_source/

$ ls t_target
config.json
```

Now, I wanted to change my config to instead link the file directly (`config.json`) instead of the directory (`t_source`).

```diff
    - link:
-       t_target: t_source
+       t_target/config.json: t_source/config.json
````

Here is where the issue lies. Remember:

1. `t_target` is a symlink pointing to `t_source`
2. `t_target/config.json` **exists** since we previously linked it.

So when dotbot goes to link this file, it sees that `t_target/config.json` exists and **deletes** it to make room for the new link. Since this points to the real original file, `t_source/config.json` is deleted!

```
$ ls t_source/config.json
t_source/config.json

$ dotbot -c dotbot.yml
Removing t_target/config.json
Nonexistent source for t_target/config.json : /home/user/test/t_source/config.json
Some links were not successfully set up

==> Some tasks were not executed successfully

$ ls t_source/config.json
ls: cannot access 't_source/config.json': No such file or directory
```

This is fundamentally user error, but I think it's worth putting some checks around since not only did it delete my source file, but it wasn't clear why it did it until I took some time to investigate.

kurtmckee:
Fixes #360

### DIFF
--- a/src/dotbot/plugins/link.py
+++ b/src/dotbot/plugins/link.py
@@ -197,6 +197,12 @@ class Link(Plugin):
         success = True
         source = os.path.join(self._context.base_directory(canonical_path=canonical_path), source)
         fullpath = os.path.abspath(os.path.expanduser(path))
+        if self._exists(path) and not self._is_link(path) and os.path.realpath(fullpath) == source:
+            # Special case: The path is not a symlink but resolves to the source anyway.
+            # Deleting the path would actually delete the source.
+            # This may happen if a parent directory is a symlink.
+            self._log.warning(f"{path} appears to be the same file as {source}.")
+            return False
         if relative:
             source = self._relative_path(source, fullpath)
         if (self._is_link(path) and self._link_destination(path) != source) or (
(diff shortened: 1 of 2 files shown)


anishathalye:
Thank you!

PR #376, opened 2025-02-20
Add a Dependabot config to maintain GitHub action versions

kurtmckee:
If accepted, Dependabot will scan the repo for GitHub action version updates once per month and submit a single PR for any actions that have received updates.

Since the repo only pins to major versions (like `actions/checkout@v4`), this will probably be a small number of PRs per year. Nevertheless, it will help reduce maintenance burden as actions are updated and old versions are deprecated.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"


anishathalye:
We pin major versions; would Dependabot submit a PR to bump a major version? How exactly does that work, since actions' APIs can change with major version upgrades? I couldn't answer this question based on the docs I could find [[1], [2]].

[1]: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot
[2]: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates

anishathalye:
Sounds good!

PR #377, opened 2025-03-07
Add ability to pass multiple config files

alipatti:
This commit adds the ability to pass multiple config files as command line arguments:

```bash
./bin/dotbot -c config_generic.yaml -c config_platform_specific.yaml
```

The config files are first concatenated in the order they are passed, and then the rest of
the dotbot code proceeds as it already exists. This felt like the least-intrusive way to
implement the feature, but happy to modify it if you see a better path.

Thanks! Awesome project you have here!

### DIFF
--- a/src/dotbot/cli.py
+++ b/src/dotbot/cli.py
@@ -3,7 +3,7 @@ import os
 import subprocess
 import sys
 from argparse import ArgumentParser, RawTextHelpFormatter
-from typing import Any
+from typing import Any, List
 
 import dotbot
 from dotbot.config import ConfigReader, ReadingError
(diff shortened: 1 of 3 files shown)


anishathalye:
Thanks! I tweaked the implementation a bit, switching to `nargs="+"` (to match what we do with other options that support multiple arguments, like `--only`) and improving error messages (e.g., giving a better error message rather than a stack trace when a config file is a dict instead of a list).

Squashed in 1bc9fb50e77cc42df9596bcf25973bcea63a515f and merged in a6c5f036e4050eac092c36ce5daaf23d3310b43f.

PR #383, opened 2025-09-05
Improve logging when linking fails

camflan:
Added verbose output to see the underlying OSError if linking fails


anishathalye:
Thank you! Merged in ee4518842adcb7b59f52793b710a7e3d1d05310d.

PR #397, opened 2026-04-15
Report failed action names on dispatch failure

ajay:
When dotbot fails, the error message currently only says:

```
Some tasks were not executed successfully
```

This provides no indication of *which* actions failed, forcing the user to scroll back through the log to identify the problem. This is especially painful with long-running configs that have many actions (package installs, linking, shell commands, etc.), or when running with verbose flags.

This PR adds tracking of failed action names and includes them in the error message.

**Before:**
```
[ERROR] Some tasks were not executed successfully
```

**After:**
```
[ERROR] Some tasks were not executed successfully: shell-dnf, link
```

#### Backwards Compatibility

The `dispatch() -> bool` signature is unchanged. The new `get_failed_actions()` method is additive. Existing third-party plugins that call `dispatcher.dispatch(tasks)` will continue to work without modification.

### DIFF
--- a/src/dotbot/cli.py
+++ b/src/dotbot/cli.py
@@ -131,7 +131,8 @@ def main() -> None:
         if success:
             log.info("All tasks executed successfully")
         else:
-            msg = "Some tasks were not executed successfully"
+            failed_str = ", ".join(dispatcher.get_failed_actions())
+            msg = f"Some tasks were not executed successfully: {failed_str}"
             raise DispatchError(msg)  # noqa: TRY301
     except (ReadingError, DispatchError) as e:
         log.error(str(e))  # noqa: TRY400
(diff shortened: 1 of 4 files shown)


anishathalye:
Overall seems like a fine idea. One detail is that Dotbot configuration is a list of "tasks", where each task can have multiple "actions". This is mostly historical, I think these days most people use one action per task.

What should the right behavior be when there are potentially multiple actions per task, and the same action can appear multiple times across tasks (e.g., there may be multiple `shell`s)? Should we show the unique set of actions that failed? Show failed actions with multiplicity?

Also, to avoid mixing up tasks and actions in the error message, we could make the error message something like:

```
Some tasks were not executed successfully (failed actions: shell-dnf, link)
```

Thoughts?

